Issue #1, opened 2020-03-02
Finish splitting API from main stac-spec


matthewhanson:
This repo contains the contents of the api-spec folder from stac-spec (0.9.0), but there are numerous links and references that need to be updated.

Additional text should be added that this is the API spec and point to stac-spec.

One question is how do we reference individual pages in the main stac-spec?  Do we reference master from master branch and dev from dev (really annoying), or tag a specific version?

cholmes:
Didn't see this before I started linking. But I linked to master branch. I think it'd be ok to link from dev to master - the idea is that these specs now move semi-independently and cut releases at different times. 

Thinking about it that could get annoying if all 'released' versions in stac-api-spec are referring to master, as things change. I'd say the ideal would be that a released version of stac-api-spec should depend on a specific version of stac-spec, and ideally we'd update the links to point at those. That sounds like it could be annoying, though maybe that should just spur us to limit the number of links.

cholmes:
Copied from https://github.com/radiantearth/stac-spec/blob/v0.9.0/CONTRIBUTING.md and made some tweaks.

m-mohr:
Anything else we need to do for this issue in the API repo?

PR #2, opened 2020-03-03
Updates for stac-api-spec repo to be more standalone

cholmes:
**Related Issue(s):** #1 

**Proposed Changes:**

*Various Readme changes:*
* Fixed links
* Added logo, 'warning', contributing and communication sections
* Refined 'about' section
* Various tweaks to represent separate repositories

**PR Checklist:**

- [ ] I have added my changes to the [CHANGELOG](https://github.com/radiantearth/stac-spec/blob/dev/CHANGELOG.md) **or** a CHANGELOG entry is not required.
- [ ] API only: I have run `npm run generate-all` to update the [generated OpenAPI files](https://github.com/radiantearth/stac-spec/blob/dev/api-spec/README.md#openapi-definitions).


api-spec.md
```diff
@@ -1,26 +1,28 @@
...
+be queried ([7.11](http://docs.opengeospatial.org/is/17-069r3/17-069r3.html#_collections_)), and each describes basic 
+information about the geospatial dataset, like its name and description, as well as the spatial and temporal extents of all 
+the data contained. [STAC collections](https://github.com/radiantearth/stac-spec/collection-spec/README.md) contain this same 
+information, along with other STAC specific fields to provide additional metadata for searching spatiotemporal assets, and 
+thus are compliant with both OAFeat Collections and STAC Collections and are returned from the `/collections/{collection_id}` 
+endpoint.
 
 In OAFeat Features are the individual records within a Collection and are provided in GeoJSON format. 
-[STAC Items](../item-spec/README.md) are analogous to OAFeat Features, are in GeoJSON, and are returned from the 
+[STAC Items](https://github.com/radiantearth/stac-spec/item-spec/README.md) are analogous to OAFeat Features, are in GeoJSON, and are returned from the
```
m-mohr: @cholmes Those kind of links are all invalid... they need to include the version/tree, but I feel like that's quite annoying to update. Wondering whether we should include the stac-spec as Git submodule here? Then it behaves like a subfolder, links are simple and link to the right API version. What do you think @matthewhanson ?

cholmes: Sub-module would definitely be interesting, especially if we can tie that to a specific version. I know these ones need to change, but was thinking to just do master -> master until we do the first release, and then start tieing them to specific versions, since structures may change. So I'd lean towards using master and then just file an issue that they need to be updated to tie to a version.

m-mohr: I could do the sub-module thing... It's fixed to a commit (i.e. version). All other options seem like duplicated work

cholmes:
Awesome, let's do it.

PR #21, opened 2020-04-27
cleanup Context Extension text

philvarner:
**Related Issue(s):** n/a

**Proposed Changes:**

1. Remove the non-existent "next" parameter as required in the yaml spec
2. Reorder properties so they're always in order "returned, limit, matched"

**PR Checklist:**

- [X] CHANGELOG entry is not required.
- [X] API only: I have run `npm run generate-all` to update the [generated OpenAPI files](https://github.com/radiantearth/stac-api-spec/blob/dev/README.md#openapi-definitions).

### DIFF
--- a/STAC-extensions.yaml
+++ b/STAC-extensions.yaml
@@ -1644,9 +1644,12 @@ components:
         context:
           type: object
           required:
-            - next
             - returned
           properties:
+            returned:
+              type: integer
+              minimum: 0
+              example: 1
             limit:
               type: integer
               nullable: true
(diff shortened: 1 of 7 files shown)


extensions/context/examples/example.json
```diff
@@ -1,9 +1,55 @@
 {
   "type": "FeatureCollection",
-  "features": [ ],
   "context": {
-    "limit": 10,
-    "matched": 1092873,
-    "returned": 9
-  }
+    "returned": 1,
```
philvarner, line 4: this example is not correct, since the "returned" value matches the number of features.

STAC-extensions.yaml
```diff
@@ -1644,9 +1644,12 @@ components:
         context:
           type: object
           required:
-            - next
```
philvarner, line 1647: removed, as it is no longer a field at all

philvarner:
> Seems to be mostly fixes, but maybe still add a CHANGELOG entry?
@m-mohr added

PR #41, opened 2020-09-14
specify type of search Link, wordsmith other API text

philvarner:
**Related Issue(s):** #

n/a

**Proposed Changes:**

1. require root link for search be `application/geo+json`, which matches the response type of the /search endpoint

**PR Checklist:**

- [ ] n/a API only: I have run `npm run generate-all` to update the [generated OpenAPI files](https://github.com/radiantearth/stac-api-spec/blob/dev/README.md#openapi-definitions).

### DIFF
--- a/api-spec.md
+++ b/api-spec.md
@@ -104,14 +104,19 @@ See the [OpenAPI specification document](openapi/STAC.yaml).
 | /search       | [ItemCollection](../item-spec/itemcollection-spec.md) | Retrieves a group of Items matching the 
 provided search predicates, probably containing search metadata from the `search` extension |
 
-The root endpoint (`/`) should function as a complete `Catalog` representation of all the data contained in the API. All `Collections` and `Items` should be linked to from the root through a `data` Link (returning all `Collections`), one or more `child` Links each referencing a single `Collection`, and a `search` Link referencing the `/search` endpoint.
+The root endpoint (`/`) is most useful when it presents a complete `Catalog` representation of all the data contained in the API, such that all `Collections` and `Items` can be navigated to by transitively traversing links from this root. This spec does not require any API endpoints from OAFeat or STAC API to be implemented, so these links may not exist if the endpoint has not been implemented.
+
+Links with these `rel` attributes should exist in the root endpoint if the reference API endpoint is implemented:
+* `data` with href to the OAFeat `/collections` endpoint
+* `child` (one or more) with href to a single `Collection` at the OAFeat `/collections/{collectionId}` endpoint
+* `search` with href to the `/search` endpoint (**required** if search endpoint is implemented)
 
 The `/search` endpoint is similar to the `/collections/{collectionId}/items` endpoint in OAFeat in that it 
 accepts parameters for filtering; however, it performs the filtering across all collections. The parameters accepted are 
 the same as the Filter Parameters above; however, the *[extensions](extensions/README.md)* also provide advanced querying 
 parameters.
 
-If the `/search` endpoint is implemented, it is recommended to add a Link to the root endpoint (`/`) with the `rel` type set to `search` that refers to the search endpoint in the `href` property, with a `type` of `application/geo+json`. This Link should look like:
+If the `/search` endpoint is implemented, it is **required** to add a Link to the root endpoint (`/`) with the `rel` type set to `search` that refers to the search endpoint in the `href` property, with a `type` of `application/geo+json`. This Link should look like:
 ```
 {
     "href": "https://example.com/search",
--- a/api-spec.md
+++ b/api-spec.md
@@ -104,12 +104,11 @@ See the [OpenAPI specification document](openapi/STAC.yaml).
 | /search       | [ItemCollection](../item-spec/itemcollection-spec.md) | Retrieves a group of Items matching the 
 provided search predicates, probably containing search metadata from the `search` extension |
 
-The `/` endpoint should function as a complete `Catalog` representation of all the data contained in the API and linked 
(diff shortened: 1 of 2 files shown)


cholmes (approved):

api-spec.md
```diff
@@ -105,15 +105,22 @@ See the [OpenAPI specification document](openapi/STAC.yaml).
 provided search predicates, probably containing search metadata from the `search` extension |
 
 The `/` endpoint should function as a complete `Catalog` representation of all the data contained in the API and linked 
-to in some way from root through `Collections` and `Items`.
+to in some way from root to `Collections` and `Items` through `children` and `child` Links.
```
m-mohr, line 108: What is a `children` link? Was that meant to be:

```suggestion
to in some way from root to `Collections` and `Items` through `child` and `item` Links.
```

api-spec.md
```diff
@@ -105,15 +105,22 @@ See the [OpenAPI specification document](openapi/STAC.yaml).
...
 
 The `/search` endpoint is similar to the `/collections/{collectionId}/items` endpoint in OGC API - Features in that it 
 accepts parameters for filtering, however it performs the filtering across all collections. The parameters accepted are 
 the same as the Filter Parameters above, however the *[extensions](extensions/README.md)* also provide advanced querying 
 parameters.
 
-If the `/search` endpoint is implemented, it is **required** to add a link with the `rel` type set to `search` to the 
-`links` array in `/` that refers to the search endpoint in the `href` property.
+It is **required** to add to the root endpoint (`/`) object a Link in the `links` array with the `rel` type set to `search`  that refers to the search endpoint in the `href` property, with a `type` of `application/geo+json`. This Link should look like:
```
m-mohr, line 115: Why was the constraint: "If the `/search` endpoint is implemented," removed? We don't implement search for example and would not be STAC API compliant anymore.

m-mohr (changes requested):
Some minor things in the comments.

api-spec.md
philvarner: I see now that `children` is not one of the Relation types listed in https://github.com/radiantearth/stac-spec/blob/master/catalog-spec/catalog-spec.md -- maybe it should be? I copied `children` from https://earth-search.aws.element84.com/v0 when updating EOD, where the href is to `/collection`.  @matthewhanson maybe you have a view on this?

philvarner: if we don't add `children`, I would phrase as:

The `/` endpoint should function as a complete `Catalog` representation of all the data contained in the API. This data is linked to from the root by a `search` Link and one or more `child` Links.

m-mohr: We have child defined for Catalog/Collections. What's wrong with that?
The reference to /collections is `data` as defined in OAFeat.

philvarner: I wasn't aware that's rel=data in OAFeat.

api-spec.md
philvarner: So you aren't required to implement `/search` in order to be STAC API compliant?  What is the minimum you're required to implement to be considered STAC API compliant?

philvarner: I removed this because I though it was supposed to be "when" rather than "if"

m-mohr: I don't know.

OAFeat requires /, /conformance, /collections and /collections/:id, I think.

Anyway, this change is basically a breaking change and is not part of the CHANGELOG, but I'd prefer to also call a OAFeat API returning STAC Items/Collections a STAC API although it doesn't implement search.

philvarner: Okay, re-added that language. Also changed it so that adding the root link to search is recommended instead of required -- we should have a larger discussion around what exactly the minimum requirement should be

m-mohr: Why have you changed it to a recommendation?

philvarner: I'm fine with that. I forgot the "required" was there already -- I thought I'd added that.  I didn't include it because I thought that it was odd to say that implementing the /search endpoint is optional, but adding it to Links is required.

api-spec.md
```diff
@@ -104,16 +104,22 @@ See the [OpenAPI specification document](openapi/STAC.yaml).
 | /search       | [ItemCollection](../item-spec/itemcollection-spec.md) | Retrieves a group of Items matching the 
 provided search predicates, probably containing search metadata from the `search` extension |
 
-The `/` endpoint should function as a complete `Catalog` representation of all the data contained in the API and linked 
-to in some way from root through `Collections` and `Items`.
+The root endpoint (`/`) should function as a complete `Catalog` representation of all the data contained in the API. All `Collections` and `Items` should be linked to from the root through a `data` Link (returning all `Collections`), one or more `child` Links each referencing a single `Collection`, and a `search` Link referencing the `/search` endpoint.
```
m-mohr, line 107: The `data` rel seems a bit unclear. I think we should just say that data should point to /collections, right? Also, that is a requirement in OAFeat (but not required for a STAC API that only implements search). And then I'm wondering why we should link to all collections in / if there's the pointer to /collections?

philvarner: I thought this was unclear also.  Yes, the example in OAFeat is 
```
    { 
      "href": "http://data.example.org/collections",
      "rel": "data", 
"type": "application/json", 
"title": "Information about the feature collections" }
```

I'm not sure what you mean by "pointer to /collections"?

philvarner: I just reworded it -- I think it's more clear now that most of the endpoints are optional, so the links are also.

m-mohr: That reads much better, thanks!

m-mohr (approved):

Issue #33, opened 2020-07-29
Align STAC Transactions with OGC API - Features Transactions


cholmes:
STAC Transactions extension pre-dated the OGC API work on transactions. But they are very similar and the OGC API has a growing community around it, so we should just align STAC with those. See http://docs.opengeospatial.org/DRAFTS/20-002.html and https://github.com/opengeospatial/ogcapi-features/tree/master/extensions/transactions/simple

This was discussed on the STAC call 7/20/20, and agreed to be the best coarse of action.

PR #11, opened 2020-04-01
Fixes and CI for OpenAPI files 

m-mohr:
**Related Issue(s):** #8, #10

**Proposed Changes:**

*Changes based on #10, so need to merge that first.*

- [x] Enable CI for OpenAPI files
- [x] Fix #8
- [x] Fix issues reported by Spectral (`npm run check-openapi`) 

**PR Checklist:**

- [x] This PR has **no** breaking changes.
- [x] I have added my changes to the [CHANGELOG](https://github.com/radiantearth/stac-api-spec/blob/dev/CHANGELOG.md) **or** a CHANGELOG entry is not required.
- [x] API only: I have run `npm run generate-all` to update the [generated OpenAPI files](https://github.com/radiantearth/stac-api-spec/blob/dev/README.md#openapi-definitions).

PR #48, opened 2020-11-02
Align with OGC API - Features Part 4: Simple Transactions extension

cholmes:
updated to make clear the relationship with ogc api - features simple transactions extension

**Related Issue(s):** #33 

**Proposed Changes:**

1. Put in links to OGC API - Features Part 4: Simple Transactions extension, and made clear that STAC's extends. Thankfully
the semantics seemed to all be the same, but we should check to be sure when they have an openapi sample document

**PR Checklist:**

- [n/a] This PR is made against the dev branch (all proposed changes except releases should be against dev, not master).

### DIFF
--- a/extensions/transaction/transaction.fragment.yaml
+++ b/extensions/transaction/transaction.fragment.yaml
@@ -7,7 +7,6 @@ paths:
       operationId: postFeature
       tags:
         - Transaction Extension
-        - insert
       parameters:
         - $ref: '#/components/parameters/collectionId'
       requestBody:
@@ -33,7 +32,7 @@ paths:
           content:
             application/json:
               schema:
-                $ref: '#/components/responses/Feature' 
+                $ref: '#/components/schemas/item'
         '400':
           $ref: '#/components/responses/BadRequest'
         '404':
@@ -66,7 +65,6 @@ paths:
       operationId: updateFeature
       tags:
         - Transaction Extension
-        - update
       parameters:
         - $ref: '#/components/parameters/collectionId'
         - $ref: '#/components/parameters/featureId'
@@ -148,7 +146,6 @@ paths:
       operationId: deleteFeature
       tags:
         - Transaction Extension
-        - delete
       parameters:
         - $ref: '#/components/parameters/collectionId'
         - $ref: '#/components/parameters/featureId'
--- a/extensions/transaction/transaction.fragment.yaml
+++ b/extensions/transaction/transaction.fragment.yaml
@@ -7,6 +7,7 @@ paths:
       operationId: postFeature
       tags:
         - Transaction Extension
+        - insert
       parameters:
         - $ref: '#/components/parameters/collectionId'
       requestBody:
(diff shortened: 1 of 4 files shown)


m-mohr:
Should we convert this PR to a draft until we have the OpenAPI document from OAFeat?

cholmes:
Ideally, but I'm not sure if we'll get updated openapi document by the time I'd like to release our beta.1

I guess I'd be inclined to put a bit more warning about that fact we are waiting for that, but that in the meantime people can just use our fragment.

Or I suppose we could attempt to contribute our fragment...

cholmes:
Actually it looks like there is an OpenAPI sample in the ogc features repo: https://github.com/opengeospatial/ogcapi-features/blob/master/extensions/transactions/simple/examples/openapi/ogcapi-features-4-example1.yaml

I think we don't want to put it in our OAFeat core yaml though, since then it'll get built with our STAC core, even though it's an extension. So I think it'd make sense to update our transaction extension yaml to align with theirs, and then merge it in.

cholmes:
Ok, I think this one is ready. Just worked to align our openapi with theirs. No major changes - changed our 200's to their 204's, used application/json with a ref instead of geojson and describing it, changed our 5xx to 500's.

I left reference to our Items, since they are features, and I think in ours we want to just allow STAC items, not any feature.

extensions/transaction/transaction.fragment.yaml
```diff
@@ -30,16 +31,15 @@ paths:
...
               description: A string to ensure the item has not been modified
           content:
-            application/geo+json:
-              schema:
-                type: string
-            text/html:
+            application/json:
               schema:
-                type: string
+                $ref: '#/components/responses/Feature'
```
matthewhanson, line 36: This is causing an error in Circle:
```
167:23    error  invalid-ref                    '#/components/responses/Feature' does not exist
```

because Feature is not defined in the components->responses section (around line 228)

matthewhanson (changes requested):

extensions/transaction/transaction.fragment.yaml
m-mohr: This has been introduced with #11 - meabe we should actually merge OAFeat.yaml and STAC.yaml? Or fix the merge tool?

cholmes: I'm fine with either route. Though I'm not sure how to fix the merge tool. And not sure exactly how to merge them by hand, do you have to line up the sections? Or just put them both in one?

matthewhanson: I think manual merge is probably the fastest and easiest, not worth messing with the merge tool for a one-time merge that isn't going to change much between versions. 
We'd have to combine like sections.

cholmes: Ok. If you can take a crack at that then go for it Matt. If not I'll take it after I do compliance/item collection / relative links

m-mohr: I'm not sure it's a good idea to just merge them as they are. I think we should try to already split them according to the conformance classes and leave the OAF schema as untouched as possible to help with future migrations. Still, it should somehow reflect STAC of course.

cholmes: I'm leaning towards just reverting this change, and go back to a string and just not have it fully aligned. Or could also just leave this PR out entirely of beta.1 - just put in a note that they are very similar and we are working to fully align.

m-mohr: It was easier than expected, changed a ref and that was basically it. Not sure everything is 100% correct, but should be better than before...

matthewhanson (approved):

m-mohr:
Was still reviewing it, but seems fine. Could probably be aligned a bit more and eventually replace our OpenAPI mostly, but we can do that later.

cholmes:
Ah, apologies - should have looked for your approval. But yeah, I think we can iterate and align more, especially when they go to final. But this will get it on people's STAC radar.

Issue #47, opened 2020-10-26
Clarify BBOX format


cholmes:
From @jjrom: 'The format of bbox is unclear from the specification. The astrea implementation waits for a comma separated delimiter string (i.e. lon1,lat1,lon2,lat2) while the Element84 implementation waits for an array i.e. ([lon1,lat1,lon2,lat2])'

He found that Features API is without an array, since we aiming to fully align with that I think we should use that, and make it clear.

jisantuc:
fwiw [Franklin](https://github.com/azavea/franklin/blob/300cbf01cbf3062dad9580f3b52fb15ca3c7b73f/application/src/main/scala/com/azavea/franklin/api/schemas/package.scala#L40-L55) also assumes no array brackets for bboxes in query params

jjrom:
Same for https://github.com/jjrom/resto - it also assumes no array brackes for bboxes in query params

cholmes:
Am making a PR for this and it seems pretty clear. Will put in a few more warnings, but I think we mostly need to get @matthewhanson to update the element84 service.

Issue #28, opened 2020-06-12
Don't allow 5 elements in bbox


m-mohr:
Don't allow 5 elements in bboxes.

Implement upstream changes from
* https://github.com/opengeospatial/ogcapi-features/issues/401
* https://github.com/radiantearth/stac-spec/pull/844

Issue #4, opened 2020-02-07
Add x-www-form-urlencoded recommendation


cholmes:
In radiantearth/stac-spec#724 we agreed that it would be good to have `x-www-form-urlencoded` as a recommendation. But we decided to take it out for RC2 because it is not specified anywhere, and has no explanation as to why someone might use it.

With that text it should easily get back into the spec, but we want to get rc2 / 0.9.0 out asap.

matthewhanson:
Talked about this in sprint #6, not convinced we really need to do this, but waiting for @philvarner to weigh in.

m-mohr:
Closing as stale issue for now.

PR #50, opened 2020-11-02
Bbox improvements

cholmes:
**Related Issue(s):** #28 #47 

**Proposed Changes:**

1. Update the openapi yaml to only allow 4 or 6 element bbox arrays
2. Link to examples and put in a bit more language around bbox

**PR Checklist:**

### DIFF
--- a/STAC-extensions.yaml
+++ b/STAC-extensions.yaml
@@ -527,15 +527,13 @@ components:
       required: false
       schema:
         type: array
-        minItems: 4
-        maxItems: 6
-        items:
-          type: number
         oneOf:
           - minItems: 4
             maxItems: 4
           - minItems: 6
             maxItems: 6
+        items:
+          type: number
       style: form
       explode: false
     collectionId:
--- a/STAC-extensions.yaml
+++ b/STAC-extensions.yaml
@@ -531,6 +531,11 @@ components:
         maxItems: 6
         items:
           type: number
+        oneOf:
+          - minItems: 4
+            maxItems: 4
+          - minItems: 6
+            maxItems: 6
       style: form
       explode: false
     collectionId:
--- a/api-spec.md
+++ b/api-spec.md
@@ -210,7 +210,6 @@ GET /search?collections=landsat8,sentinel&bbox=-10.415,36.066,3.779,44.213&limit
 }
 ```
 
-
 | Parameter    | Type             | APIs         | Description |
 | -----------  | ---------------- | ------------ | ----------- |
 | collectionId | string           | OAFeat       | **Path-only** Single Collection ID to include in the search for items. Only Items in one of the provided Collection will be searched |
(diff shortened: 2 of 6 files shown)


STAC-extensions.yaml
```diff
@@ -531,6 +531,11 @@ components:
         maxItems: 6
```
m-mohr, line 531: If you add the oneOf, you need this line and the one above.

m-mohr, line 531: This is strange, there's still the old min/maxItems in the file, in additon to the new oneOf. But it should not. Does that get included from the OAFeat part? Should be fixed...

api-spec.md
```diff
@@ -135,7 +135,7 @@ This Link should look like:
 
 Unless otherwise noted by **Path-only**, these filters are passed as query string parameters, form parameters, or JSON 
 entity fields.  For filters that represent a set of values, query and form parameters should use comma-separated 
-string values and JSON entity attributes should use JSON Arrays. 
+string values (with no outer brackets \[ or \]) and JSON entity attributes should use JSON Arrays.
```
m-mohr, line 138: I'm confused. So it's comma separated values, but "JSON entity attributes should use JSON Arrays"? Does that conflict? What does "JSON entity attributes should use JSON Arrays" actually mean?

m-mohr (changes requested):

STAC-extensions.yaml
cholmes: Ugh, I thought this would be an easy fix. Good catch on the OAFeat part. I updated it but forgot to run the generation stuff. So I did, and now have:

```
        type: array
        oneOf:
          - minItems: 4
            maxItems: 4
          - minItems: 6
            maxItems: 6
          - minItems: 4
            maxItems: 4
          - minItems: 6
            maxItems: 6
```

Not helpful. Going to try to dig into how merging works... Any ideas are welcome.

api-spec.md
cholmes: I'm 90% sure that this means:
- If you are using query string parameters (GET request to `/search/`) you should do bbox=10,10,20,20
- If you are using JSON requests (POST to `/search/`) then you should do "bbox": [10,10,20,20]

But obviously, that is not clear. If my explanation makes sense then I can expand this description out. It occurs to me we should also just remove the 'form parameters', and perhaps put it back in if we do https://github.com/radiantearth/stac-api-spec/issues/4

STAC-extensions.yaml
cholmes: Ok, ended up just removing most of the BBOX elements from STAC.yaml, as it says we extend the OAFeat one, but then we replicate a number of fields. So made it so it was just the description, where we add an example. 

Hopefully everything is right now.

api-spec.md
m-mohr: Oh, I see. Let's better fix that, too, but I'm approving anyway for now as the main issue has been solved.

m-mohr (approved):

Issue #32, opened 2020-07-29
Align STAC Filtering with OGC CQL


cholmes:
STAC wrote its own filter language, since OGC API - Features had not established one. But the OGC community is now aligning around CQL, see http://docs.opengeospatial.org/DRAFTS/19-079.pdf and https://github.com/opengeospatial/ogcapi-features/tree/master/extensions/cql/

On the STAC Call of 7/20/20 everyone agreed that we should aim to drop our custom language and adopt CQL. 

We should redo the specification to use CQL exactly, just as the features API does. And for STAC we want to require the use of CQL, so STAC is a profile on top of Features API that requires a bit more. Users who want to create a STAC implementation without supporting filtering can just implement STAC core (and point a more capable STAC API at it to ingest).

cholmes:
Moving to beta.2, as OGC CQL is a bit unstable beneath us, see https://github.com/radiantearth/stac-api-spec/pull/51#issuecomment-736696528

dwilson1988:
@cholmes 
Hey, I'm in the process of watching this moving target. Currently working on our next generation of STAC API (implemented in Go). Currently using the query extension as it stands in the repo, but while we are upgrading, this appears to be a good time to align with CQL.  

I haven't touched CQL since my days working with Geoserver, it looks like it's changed up a bit from what I remember. 

From what I can tell individual JSON filters are AND'd and OR'd together with a literal? Is that correct? Does STAC plan to do something a little more clear for the query extension, such as say {"and": [<filter_1>, <filter_2>]}? Will it be in the query extension or mirror OAFeat directly? I'm curious on the progress with this. Even if it's not fully settled, I would like to start an implementation on our end.

cholmes:
Hey @dwilson1988 - the plan for the next STAC release is to fully align with the latest OGC CQL. They made some tweaks to their JSON since I wrote the PR. There are a couple of STAC implementations right now investigating CQL. The plan is to deprecate our STAC 'query' extension, since it never really got to fruition, and just fully adopt the OGC CQL (using the filter keyword). So yes, we'll just mirror OAFeat directly.

[...]

PR #51, opened 2020-11-02
Align to OGC CQL

cholmes:
**Related Issue(s):** #32 

**Proposed Changes:**

1. Deprecated 'query' and made a new extension called 'filter' which references the OGC CQL document.
2. 

**PR Checklist:**


cholmes:
TODO: Update examples.md and any other examples to use CQL
TODO: Figure out what to do about 'intersects' in STAC - deprecate? Or leave as a short cut?

cholmes:
Ok, I think this is ready for review. It's the biggest PR for the release, so I may have forgotten some bits, so review is appreciated @m-mohr and @matthewhanson 

I think this process got me to figure out openapi more, and I ran it through the stac `npm run serve-ext` and it seemed sensible. 

I marked the query extension as 'deprecated', but then I did remove it from the openapi build, since it seemed like it'd be confusing to have it in there too. I suppose I could update its openapi document to say 'deprecated' everywhere if we want, but I feel like there's no reason to confuse new users, and implementors of older versions will figure out that it's deprecated.

STAC-extensions.yaml
```diff
@@ -337,7 +337,7 @@ paths:
         - $ref: '#/components/parameters/limit'
         - $ref: '#/components/parameters/ids'
         - $ref: '#/components/parameters/collectionsArray'
-        - $ref: '#/components/parameters/query'
```
matthewhanson, line 340: If this is deprecated rather than removed shouldn't it remain in the OpenAPI doc for now?

STAC-extensions.yaml
```diff
@@ -629,15 +629,28 @@ components:
...
+    filter-lang:
+      name: filter-lang
+      in: query
+      required: false
+      schema:
+        type: string
+        enum:
+          - cql-text
+          - cql-json
+        default: cql-text
```
matthewhanson, line 652: For POST should it default to `cql-json`?

cholmes: I thought about it, and even had a note in the text. But wasn't sure how to even go about specifying it, so ended up just dropping it in favor of using the OGC API examples directly, and figured people could be explicit about it. If you have a good openapi way to add it that'd be great, it's beyond my current openapi skills. We could also just open an issue for it, in our repo and probably in OGC API CQL as well.

STAC-extensions.yaml
cholmes: I wasn't sure. But yeah, happy to leave it in. Is there a way to mark things as deprecated in openapi?

m-mohr: Yes, OpenAPI has a deprecated flag you can set in some contexts.

STAC-extensions.yaml
m-mohr: There's surely a way to do that. Likely remove the default value here and then add it back again where the refs for this schema are with allOf.

api-spec.md
```diff
@@ -87,12 +88,10 @@ operation to create an Item. Any query against a single OAFeat collection endpoi
 the collection name in the `collections` query parameter.
 
 It is recommended for clients use POST for querying (if the STAC API supports it), especially when using the 
-`intersects` query parameter, for two reasons:
+`intersects` query parameter, for one main reason:
 
 1. In practice, the allowed size for an HTTP GET request is significantly less than that allowed for a POST request,
```
matthewhanson, line 93: no need to number if there's just one reason

api-spec.md
```diff
@@ -276,14 +276,13 @@ These parameters and fields are reserved for the Sort extension.
 | --------- | ----------------- | ----------- |
 | sort      | string \| \[Sort] | Placeholder parameter for [API Sort Extension](extensions/sort/README.md). |
 
-### Query Extension
+### Filter Extension
```
matthewhanson, line 279: Might want to keep Query Extension description here as well, but marked as deprecated

examples.md
```diff
@@ -43,21 +43,26 @@ Request 100 results in `mycollection` that is in New Zealand at anytime on Janua
...
-        "type": "Polygon",
-        "coordinates": [[
-            [-77.0824, 38.7886], [-77.0189, 38.7886],
-            [-77.0189, 38.8351], [-77.0824, 38.8351],
-            [-77.0824, 38.7886]
-        ]]
-    },
+    "filter": {
+        "intersects": {
+                "property": "geometry",
```
matthewhanson, line 55: will this be required? since it's GeoJSON it will always be geometry

extensions/filter/README.md
```diff
@@ -0,0 +1,118 @@
...
+
+## Example
+
+Before we dive deep into an overview of CQL, it is worth showing a concrete examples. A common query of STAC is to find items (implementing the EO 
+extension) with cloud cover between 0 and 10%. This would look like: 
+
+### cql-text (GET)
+
+```http
+/search?filter=eo:cloud_cover < 10
```
matthewhanson, line 27: With CQL you can query top-level fields, such as geometry with the intersects parameter. Do other filters all assume they are property names?
How would one query another top-level field like `assets` with CQL?
Or do normal properties have to be qualified, e.g., "properties.eo:cloud_cover"?

matthewhanson:
Finished reviewing, left some comments. Biggest thing is leaving in info on query filter but marking as deprecated.  
Also, wondering about CQL and referencing of properties vs top-level fields. `intersects` filter specifies a property of `geometry` which suggests that actual Item properties need to be fully qualified with a `properties.` prefix.

api-spec.md
cholmes: Fair, the second reason was no longer valid. Will update.

examples.md
cholmes: I'm pretty sure it will be. Even with GeoJSON you could have another property that is a geometry I believe. We can try to get them to change it in CQL, but I'm not sure if it makes sense to.

extensions/filter/README.md
cholmes: I'm pretty sure CQL assumes the other filters all assume they are property names. With OGC I think it follows the 'feature model', which says a geometry and zero or more attributes - which in GeoJSON map to properties. But it's a good question - would be good for CQL to clarify its use with GeoJSON.

api-spec.md
cholmes: added a note here.

STAC-extensions.yaml
cholmes: Ok, marked as deprecated, added back in.

STAC-extensions.yaml
cholmes: I've been spending way too long on this release, which I mostly just wanted to do to be able to move on to stac-spec 1.0 releases. This type of openapi tweaking takes me hours, since I don't really understand it all, so I'm just going to move this to an issue and do in beta.2 unless someone is able to volunteer to fix it.

m-mohr: I've fixed the issue mentioned above and some other things that came to my attention, also tried to directly link to their CQL OpenAPI fragment. Should be fine for a first release, but will need some work in future betas.

m-mohr: By the way, they are still changing the JSON structure quite a bit (see https://github.com/opengeospatial/ogcapi-features/issues/460), so a release of CQL with STAC could lead to quite some issues?!

cholmes: Yeah, I saw that :( I had thought it was a bit more solid. Also noticed that Clemens is still working on his full review: https://github.com/opengeospatial/ogcapi-features/pull/456

Unless they get 460 merged in the next day or so then I think we should release and leave out CQL.

m-mohr (changes requested):
We need to wait for https://github.com/opengeospatial/ogcapi-features/pull/461 to be merged and then check/update examples and OpenAPI $refs.

cholmes:
Looks like they will not get this in this week - https://github.com/opengeospatial/ogcapi-features/issues/460#issuecomment-736673402

So I think we should just save this PR for beta.2, but mention that it's coming in a blog post.

This one is also going to need be reworked with #60 landing.

cholmes:
Just wanted to put a note on the status of this:

* This will likely be the main set of work for stac-api-spec 1.0.0-beta.2, as there is consensus that it is better to use an accepted standard than to try to create our own query language (which has a lot more to do if it is to be complete). So the current plan is to deprecate 'query' in beta.2, and to have 'filter' be an option, which will just directly use OGC's CQL conformance classes.
* There are several server implementations trying out CQL now. The comment period for Filter/CQL is being extended, so we should be sure to try it out soon and give our feedback. In comment period the OAFeat group is quite open to changes, especially ones driven by real world experience. So let's do our best to help evolve the core spec to meet our needs.
* The one major thing I left out that looks to be required now is 'queryables'. In general implementors looking to try out CQL and give feedback should start with the core Filter/CQL spec, and make it work with the OAFeat part of STAC. The info in this PR is likely a bit out of date (both from stac-api-spec and CQL perspectives), but it maps the general approach.

philvarner:
I created a new branch and PR https://github.com/radiantearth/stac-api-spec/pull/128 , since this one is against master and there was a large amount of rework to reapply it to develop.

extensions/filter/README.md
philvarner: to follow up on this, Part 3 requires that only queryables be used in expressions, so the implementer defines what the queryable-to-STAC field mapping is. 

I need to dig more into how (or if) one can do queries like "item has asset with band where common_band red", since `assets` is an array and `eo:bands` is an array.

cholmes:
Closing this in favor #128

Issue #42, opened 2020-09-15
What is the minimum implementation required to be considered to be STAC API compliant?


philvarner:
What is the minimum implementation required to be considered to be STAC API compliant? We don't state explicitly one way or the other. For example, implementing `/search` is optional.

m-mohr:
I would think you either implement OAFeat and return STAC compliant responses or have / and /search implemented.

matthewhanson:
Agree it should implement OAFeat and meet all those requirements.
Then root catalog, collections, and items should all return valid STAC metadata.

Where should we add this?

m-mohr:
This has been pretty much answered by #60 - you need to implement the landing page, all other is optional and you can figure out what is supported by the conformance classes published on the landing page.

Issue #27, opened 2020-06-08
New fields: stac_api_version and stac_api_extensions


joshfix:
Looking at Staccato's root catalog (https://stac.boundlessgeo.io/), links are provided for collections, catalogs, and API endpoints, however, the top of the document displays a stac version field:

`"stac_version": "1.0.0-beta.1"`

If the API and STAC specs a split, should we introduce a `stac_api_version` field or something similar?

m-mohr:
(Transferred to api-spec as it seems to be an API issue.) I think you are right, we need stac_* fields also in stac_api_* variants, but to be honest, I don't like that too much. Then our openEO landing page looks like this:

```js
api_version: "1.0.0-rc.2", // openEO version
stac_version: "0.9.0",
stac_api_version: "0.9.0",
stac_extensions: [...],
stac_api_extensions: [...], // or do we follow OGCs conformance classes behavior?
...
```

geospatial-jeff:
I think we need either `stac_*` fields as mentioned or to remove ambiguity from the api extensions.  Paging is a perfect example.  As currently written, if a client wants to use the same pagination logic across potentially multiple api implementations it isn't really possible because (1) an api may or may not implement paging and (2) the rel type used for the previous/next page (or equivalent) may be different across implementations.  Preventing a client from consistently paginating across api implementations is a pretty big flaw in my opinion (and that is just one of the available api extensions).

m-mohr:
Another idea could be to re-use OGCs /conformances (not saying I like it), but that would be the OGC API way to specify the version. They don't do it in the landing page /.

> (1) an api may or may not implement paging and

It would indeed be useful to specify what extensions are implemented.

> (2) the rel type used for the previous/next page (or equivalent) may be different across implementations

I don't understand this. It should always be next for next page. If there's a difference it seems like a wrong implementation?! Can you give an example?

> [Here](https://github.com/developmentseed/cogeo-mosaic/blob/master/cogeo_mosaic/backends/stac.py#L138-L192) is a good example of a client having to deal with said ambiguity.

What exactly is the issue worked around there? I'm not used to reading Python...

geospatial-jeff:
> Another idea could be to re-use OGCs /conformances (not saying I like it), but that would be the OGC API way to specify the version. They don't do it in the landing page /.


I think this is fine, isn't this the purpose of a conformance page?  I would also be fine with a `stac_api_extensions` field at the top level of the landing page.


[...]

PR #58, opened 2020-11-13
Minor clean-up and more intro

cholmes:
**Related Issue(s):** #42 

**Proposed Changes:**

1. Added an intro to ease people in a bit more gently
2. added paragraph on the compliance requirements, as requested in #42

**PR Checklist:**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## \[Unreleased\]
 
 ### Added
- - Clearer language on compliance requirements
+- Clearer language on compliance requirements
 
 ### Changed
 - Updated to STAC specification version 1.0.0-beta.2
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## \[Unreleased\]
 
 ### Added
-- Clearer language on compliance requirements
+- More intro text
 
 ### Changed
 - Updated to STAC specification version 1.0.0-beta.2
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
+ - Clearer language on compliance requirements
 
 ### Changed
 - Context Extension OpenAPI spec was updated to remove the no longer used `next` attribute
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## \[Unreleased\]
 
 ### Added
-- More intro text
 
 ### Changed
 - Updated to STAC specification version 1.0.0-beta.2
--- a/api-spec.md
+++ b/api-spec.md
@@ -32,7 +32,8 @@ thus are compliant with both OAFeat Collections and STAC Collections and are ret
 endpoint.
 
 In OAFeat Features are the individual records within a Collection and are usually provided in GeoJSON format. 
-[STAC Items](https://github.com/radiantearth/stac-spec/item-spec/README.md) are compliant with the OAFeat Features [GeoJSON requirements class](http://docs.ogc.org/is/17-069r3/17-069r3.html#_requirements_class_geojson),  and are returned from the 
+[STAC Items](https://github.com/radiantearth/stac-spec/item-spec/README.md) are compliant with the OAFeat Features 
+[GeoJSON requirements class](http://docs.ogc.org/is/17-069r3/17-069r3.html#_requirements_class_geojson), and are returned from the 
 `/collections/{collection_id}/items/{item_id}` endpoint. The return of other encodings 
 ([html](http://docs.ogc.org/is/17-069r3/17-069r3.html#rc_html), [gml](http://docs.ogc.org/is/17-069r3/17-069r3.html#rc_gmlsf0))
 is outside the scope of STAC API, as the [STAC Item](https://github.com/radiantearth/stac-spec/blob/master/item-spec/item-spec.md) is
--- a/api-spec.md
+++ b/api-spec.md
@@ -226,7 +226,7 @@ with a single value array in JSON if the same functionality is desired. For filt
 query parameters should use comma-separated string values (with no outer brackets \[ or \]) and JSON entity 
 attributes should use JSON Arrays. 
 
-### Examples
+### Query Examples
 
 ```http
 GET /search?collections=landsat8,sentinel&bbox=-10.415,36.066,3.779,44.213&limit=200&datetime=2017-05-05T00:00:00Z
@@ -240,6 +240,8 @@ GET /search?collections=landsat8,sentinel&bbox=-10.415,36.066,3.779,44.213&limit
     "datetime": "2017-05-05T00:00:00Z"
 }
 ```
+
+### Query Parameter Table
 
 | Parameter    | Type             | APIs         | Description |
 | -----------  | ---------------- | ------------ | ----------- |
--- a/api-spec.md
+++ b/api-spec.md
@@ -32,8 +32,7 @@ thus are compliant with both OAFeat Collections and STAC Collections and are ret
 endpoint.
 
 In OAFeat Features are the individual records within a Collection and are usually provided in GeoJSON format. 
-[STAC Items](https://github.com/radiantearth/stac-spec/item-spec/README.md) are compliant with the OAFeat Features [GeoJSON
-requirements class](http://docs.ogc.org/is/17-069r3/17-069r3.html#_requirements_class_geojson),  and are returned from the 
+[STAC Items](https://github.com/radiantearth/stac-spec/item-spec/README.md) are compliant with the OAFeat Features [GeoJSON requirements class](http://docs.ogc.org/is/17-069r3/17-069r3.html#_requirements_class_geojson),  and are returned from the 
 `/collections/{collection_id}/items/{item_id}` endpoint. The return of other encodings 
 ([html](http://docs.ogc.org/is/17-069r3/17-069r3.html#rc_html), [gml](http://docs.ogc.org/is/17-069r3/17-069r3.html#rc_gmlsf0))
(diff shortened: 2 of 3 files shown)


api-spec.md
```diff
@@ -40,6 +54,23 @@ Implementations may **optionally** provide support for the full superset of STAC
 where the collection ID in the path is equivalent to providing that single value in the `collections` query parameter in 
 STAC API.
 
+### STAC API Compliance
```
m-mohr, line 57: How to expose only STAC Collections in an API?

cholmes: I'm open on what exactly we do here, was just trying to capture what was in the PR.

My thinking was that a Collection only endpoint is actually _not_ a STAC API. In my mind a STAC API is when you provide 'search'. OpenEO is a compliant STAC catalog, as is https://storage.googleapis.com/pdd-stac/disasters-0.9.0/catalog.json And I'd see implementations that are servers (RESTful API's) that only implement STAC core, and don't offer search/ or collections/items/ endpoints.

I think I agree with your point about RESTful, with no specific endpoints required, but we should be able to still specify functionality required in some way? Is OGC Features API RESTful?

m-mohr: > OpenEO is a compliant STAC catalog, as is https://storage.googleapis.com/pdd-stac/disasters-0.9.0/catalog.json

Unfortunately, it is not. We use the data rel type instead of child as it is specified in STAC API. So it's actually not a valid STAC catalog... And then /collections is also not valid STA, it's only valid in STAC API.

> I think I agree with your point about RESTful, with no specific endpoints required, but we should be able to still specify functionality required in some way?

Why would that be required? If we have conformance classes anyway, couldn't we just specify what is implemented there or just follow links? I'm not sure myself yet, but I don't like that it's obviously getting towards just STAC API for Items and it seems I need to come up with my own STAC API spec for Collections.

cholmes: Ah, that's interesting on using data rel type instead of child, I'd thought they were child. I do think that's a confusing little area.

I'm not set on not having STAC API just for items, just saying that was more of my mental model. That everyone should implement STAC Core, and that you layer on Features API / Search endpoint for filtering items. And the latter is 'STAC API'. 

But for beta.1 I prefer to not rock the boat - mostly just want to align with CQL / Transactions, clean things up, and get something for people to work with. I definitely see a beta.2, and I think the recommendations for how to implement STAC catalogs / collections is core to figure out, and this to me fits in. I think the conformance classes are probably the way to go. And I'm not set on my mental model, I can see value in having 'STAC API' specify how to implement Collections / Catalogs without search. But I also like that STAC Core *can* be implemented by a dynamic server, but that we don't need to specify that.

> Anyway, we can't represent all static STACs in the API it seems...

What does this mean?

m-mohr: > Ah, that's interesting on using data rel type instead of child, I'd thought they were child. I do think that's a confusing little area.

It makes sense, because it returns you a different JSON. Not a collection/catalog, but the response of /collections, which contains collections and links. /collections then lists the "children" instead of adding (in our case) 450+ links to the links in the landing page. It's just that this is not specified in STAC itself, but only in API (originating from Features). 

> I think the conformance classes are probably the way to go

Agreed, see #27.

> I can see value in having 'STAC API' specify how to implement Collections / Catalogs without search.

Actually, I'd rather say we should have collection search at some point and then there might be no "without search" anylonger.

> > Anyway, we can't represent all static STACs in the API it seems...
> 
> What does this mean?

If we merge this PR, you need Items in collections/catalogs, otherwise the API is not compliant although it's a valid STAC.

cholmes: Ok, cool. Now I understand. Thanks.

> Actually, I'd rather say we should have collection search at some point and then there might be no "without search" any longer.

I like that. And I think records is actually getting close to a draft, so perhaps we could try to put that in for beta.2.

I think I'll cut this PR down to just be the intro text, and then aim to do #42 with #27. Have a branch started now, will push when it's to a decent point.

matthewhanson: Catching up, think this all was largely resolved in Monday's discussion.

However, regarding @m-mohr statement about not adding child links to root catalog when you have too many collections....do we have another open discussion on this?  In stac-cmr there are a few providers that have more than 1000 collections, and currently they are all added to the root endpoint, e.g., https://cmr.uat.earthdata.nasa.gov/stac/GES_DISC

matthewhanson:
Failing CircleCI, problem with markdown somewhere, I think I see where the problem is, pushing commit

api-spec.md
```diff
@@ -21,15 +31,19 @@ information, along with other STAC specific fields to provide additional metadat
 thus are compliant with both OAFeat Collections and STAC Collections and are returned from the `/collections/{collection_id}` 
 endpoint.
 
-In OAFeat Features are the individual records within a Collection and are provided in GeoJSON format. 
-[STAC Items](https://github.com/radiantearth/stac-spec/item-spec/README.md) are analogous to OAFeat Features, 
-are in GeoJSON, and are returned from the `/collections/{collection_id}/items/{item_id}` endpoint.
+In OAFeat Features are the individual records within a Collection and are usually provided in GeoJSON format. 
+[STAC Items](https://github.com/radiantearth/stac-spec/item-spec/README.md) are compliant with the OAFeat Features [GeoJSON
+requirements class](http://docs.ogc.org/is/17-069r3/17-069r3.html#_requirements_class_geojson),  and are returned from the 
+`/collections/{collection_id}/items/{item_id}` endpoint. The return of other encodings
```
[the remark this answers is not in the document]
m-mohr: The new conformance classes would require rewriting a lot of text here anyway and I actually think this sentence here wasn't really true anyway in the past.  AFAIK there are some APIs not implementing OAFeat at all, just having /search...

api-spec.md
m-mohr: Hmm, my statement about adding child links to root is a bit biased by openEO and is also weakened by the discussion on Monday. I don't think there's a major issue with your implementation. child links at the landing page are fine, BUT I'd recommend to put them in /collections and link to it with rel type data if OAFeat is implemented anyway. Also, in an API that is not necessarily only STAC (as openEO is), it would bloat the initial request with data that is not required at that stage and thus connection times can slow down. In an API just covering STAC, I'd find it's not a major problem as it's the primary goal/content of the API.

m-mohr: I think I'd just remove the whole new chapter for now, merge this and then write-up the conformance classes. Those will basically change the whole chapter anyway.

matthewhanson: My main interest is in having the API browsable by STAC browser, so I think that will need an update to respect the `data` rel type link.

cholmes: Yes, that was my plan - remove the whole chapter on compliance, and then redo it with the conformance PR.

m-mohr: @matthewhanson I implemented the data rel type some days ago into STAC Browser: https://github.com/radiantearth/stac-browser/pull/59 Already live on stacindex.org...

api-spec.md
```diff
@@ -21,15 +31,19 @@ information, along with other STAC specific fields to provide additional metadat
...
+specified in GeoJSON.
 
 A typical OAFeat will have multiple collections, and each will just offer simple search for its particular collection at 
 `GET /collections/{collectionId}/items`.
 The main STAC endpoint specified beyond what OAFeat offers is `/search`, which offers cross-collection search. A primary
 use case of STAC is to search diverse imagery collections (like Landsat, Sentinel, MODIS) by location and cloud cover for any
-relevant image. So the ability to do searches across Collections is required, and is not yet specified by OAFeat. Due to the 
+relevant image. So the ability to do searches across Collections is very important, and is not yet specified by OAFeat. Due to the 
 limited parameter support in OGC API - Features, it is recommended to use the STAC API endpoint 
 `POST /search` for advanced queries, as it supports richer options.
```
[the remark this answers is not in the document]
cholmes: There's some chance I caught this in the CQL PR? With CQL POST is no richer than GET, they are equivalent query languages, while I think for STAC it had some more?

api-spec.md
```diff
@@ -21,15 +31,19 @@ information, along with other STAC specific fields to provide additional metadat
...
+([html](http://docs.ogc.org/is/17-069r3/17-069r3.html#rc_html), [gml](http://docs.ogc.org/is/17-069r3/17-069r3.html#rc_gmlsf0))
+is outside the scope of STAC API, as the [STAC Item](https://github.com/radiantearth/stac-spec/blob/master/item-spec/item-spec.md) is
+specified in GeoJSON.
 
 A typical OAFeat will have multiple collections, and each will just offer simple search for its particular collection at 
 `GET /collections/{collectionId}/items`.
 The main STAC endpoint specified beyond what OAFeat offers is `/search`, which offers cross-collection search. A primary
 use case of STAC is to search diverse imagery collections (like Landsat, Sentinel, MODIS) by location and cloud cover for any
-relevant image. So the ability to do searches across Collections is required, and is not yet specified by OAFeat. Due to the 
+relevant image. So the ability to do searches across Collections is very important, and is not yet specified by OAFeat. Due to the
```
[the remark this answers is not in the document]
cholmes: They have identified it as future work (in https://github.com/opengeospatial/ogcapi-features/issues/451), and reference it in Filter/CQL - http://docs.opengeospatial.org/DRAFTS/19-079.html#filter-param-multiple-collections

api-spec.md
```diff
@@ -21,15 +31,19 @@ information, along with other STAC specific fields to provide additional metadat
...
```
cholmes, line 46:
```suggestion
matching item. So the ability to do searches across Collections is very important, and is not yet specified by OAFeat. Due to the 
```

api-spec.md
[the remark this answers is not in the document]
cholmes: Ah, yeah - I wrote that before I had dug deep into the geojson requirements class. Will try to remember to rewrite this for the confromance PR.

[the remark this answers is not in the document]
cholmes: @matthewhanson - can you remove your request changes? I think I got the minor ones, and the bigger core compliance things I will do in the other PR - this one is now just about some cleaner text.

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## \[Unreleased\]
 
 ### Added
+- More intro text
```
m-mohr, line 10: I don't think this needs to be in the changelog and might even confuse people.

m-mohr (approved):

CHANGELOG.md
cholmes: Fair. I was going back and forth on it. Wouldn't have added it except I already had a line there. Will take it out.

matthewhanson (approved):

PR #59, opened 2020-11-13
Conformance

cholmes:
**Related Issue(s):** #58 #27 

**Proposed Changes:**

1. Add conformance classes
2. 

**PR Checklist:**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
+ - Clearer language on compliance requirements
 
 ### Changed
 - Context Extension OpenAPI spec was updated to remove the no longer used `next` attribute
--- a/README.md
+++ b/README.md
@@ -20,15 +20,14 @@ compatible and usable with any OGC API - Features clients. The STAC API can be t
 to search STAC Catalogs, where the features returned are STAC [Items](https://github.com/radiantearth/stac-spec/item-spec/item-spec.md), 
 that have common properties, links to their assets and geometries that represent the footprints of the geospatial assets.
 
-## WARNING
+## Stability Note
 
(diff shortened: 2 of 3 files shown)


cholmes:
@m-mohr - Ok, I'm still working through this, but was curious on feedback, as I don't think I have it right. 

I've got the three conformance classes as you laid them out. But I said STAC Collections 'depends' on OGC API - Features core. But I don't think that's right, since you don't fully implement 'core', since it has items and you don't?

I wanted to say a valid STAC is either /search or /collections. But that doesn't seem to work. So it's implement either STAC Collections or (STAC search or STAC Items)? Which seems a bit complicated to explain. 

And then I'm also not sure how to handle `/` in STAC, where that fits in. We say it 'Extends `/` from OAFeat to return a full STAC catalog.'

It feels like the easy thing may be to just drop our notion that `/search` would be implemented without implementing OGC API - Features... We could tell people if they wanted to just have a single endpoint to put all their features under a single `collections/collectionsId` endpoint.

I'll keep working away on fleshing this out...

cholmes:
Ah, interesting, looking at https://earthengine.openeo.org/v1.0 I see that you do have conformance/ and it says you implement http://www.opengis.net/spec/ogcapi-features-1/1.0/conf/core. So I guess we could say STAC Collections depends on core, and then STAC Items would depend on GeoJSON. 

Hrm, though it looks like you don't actually implement the `/service-desc` endpoint. 

Looking at all the OGC requirements again does make me want to be able to offer something simpler - like just the /search/ endpoint and /conformance with a hardcoded value. Or /collections without more overhead. 

Perhaps we have a stac-core that is the landing page (stac catalog and conformance/ (borrowed from / compatible to OAFeat) and collections/ ? With /search optional, and you can use the collectionId's to limit your search, but you don't have to implement collections/{id}/items for all collections - that'd be the stac-items class?

cholmes:
this was based on the wrong branch, just going to redo this whole PR / branch.

cholmes:
Replaced by #60

Issue #25, opened 2020-05-27
Mention ItemCollection in Markdown


matthewhanson:
ItemCollection is being removed from core, discuss if it belongs here.

geospatial-jeff:
I think it should be included in the api spec.  I can't come up with a better replacement for representing groups of items in `/collections/{collectionId}/items` and `/search` responses.  Having a feature collection just makes sense since items extend geojson features.

The most atomic response is just an array of stac items, but there is value in adding the extra top level fields to create a feature collection for client compatibility.

I also think keeping the top level response as an object is valuable for adding other fields (ex. context/paging extension).

matthewhanson:
Discussed in Sprint 6, decided we don't need both JSON schema and OpenAPI.

OpenAPI docs should define the search response, closing.

philvarner:
> Discussed in Sprint 6, decided we don't need both JSON schema and OpenAPI.
> 
> OpenAPI docs should define the search response, closing.

@matthewhanson we do need to add the markdown description of ItemCollection and the sample files, correct? I'll do this, but wanted to check before creating a PR.

philvarner:
@matthewhanson also related to this is using https://github.com/sparkgeo/stac-validator requires JSON schema definitions, and it would be useful to be able to validate API responses using this in addition to Catalogs and Items.

m-mohr:
OpenAPI files are basically JSON Schema with the nullable exception, but some JSON Schema validators even understand that. 

And I don't think we need a markdown description. It's described in the (rendered) OpenAPI Files, too. Also, I don't see why we would need it for that particular response, but not the others (e.g. /collections).

cholmes:
I'm re-opening this, since I think we need at least some reference somewhere in the markdown. There are a couple references in the API spec, and it'd be weird to dump people straight into an OpenAPI document. I think it can be a brief reference, but I'd like something to link to.

cholmes:
@m-mohr - was your thumbs down that I shouldn't include a reference in the API spec?

m-mohr:
I think I misunderstood your point. I just think the OpenAPI file is usually enough normative information and I don't like having duplicated information, the original ItemCollection file was rather long, but maybe -to make it better linkable-, add a very short chapter here? https://github.com/radiantearth/stac-api-spec/blob/master/api-spec.md#ogc-api---features-endpoints

m-mohr:
[...]

Issue #12, opened 2020-04-01
Update rendered versions on stacspec.org


m-mohr:
Update
* https://stacspec.org/STAC-api.html
* https://stacspec.org/STAC-ext-api.html

to point to this repo once STAC 1.0 is released.

cholmes:
@matthewhanson - I think you said you could take this one? 

I believe we decided that we'd make this so it'd automatically publish the openapi docs to stacspec.org (like how we do with json schemas, but published to an api directory with folders for each api version), and we'd get rid of the generated version in the repo.

m-mohr:
We should not forget to add a redirect from the current files to the new locations.

cholmes:
I'm not sure if an automatic redirect makes sense until we have the full combined rendered html, since it'd be weird to redirect to just the core spec. But I think we should update the existing one with a note at the top that points people at the new repo / release.

m-mohr:
Okay, that makes sense. I guess we should go through the repos and change the old links then...

PR #60, opened 2020-11-13
Add conformance class information

cholmes:
**Related Issue(s):** #12, #25, #27, #58

**Proposed Changes:**

1. Add conformance classes, as well as clarity on what a complaint STAC API is
2. 

**PR Checklist:**


cholmes:
@m-mohr - Ok, I'm still working through this, but was curious on feedback, as I don't think I have it right.

I've got the three conformance classes as you laid them out. But I said STAC Collections 'depends' on OGC API - Features core. But I don't think that's right, since you don't fully implement 'core', since it has items and you don't?

I wanted to say a valid STAC is either /search or /collections. But that doesn't seem to work. So it's implement either STAC Collections or (STAC search or STAC Items)? Which seems a bit complicated to explain.

And then I'm also not sure how to handle / in STAC, where that fits in. We say it 'Extends / from OAFeat to return a full STAC catalog.'

It feels like the easy thing may be to just drop our notion that /search would be implemented without implementing OGC API - Features... We could tell people if they wanted to just have a single endpoint to put all their features under a single collections/collectionsId endpoint.

I'll keep working away on fleshing this out...

cholmes:
Ah, interesting, looking at https://earthengine.openeo.org/v1.0 I see that you do have conformance/ and it says you implement http://www.opengis.net/spec/ogcapi-features-1/1.0/conf/core. So I guess we could say STAC Collections depends on core, and then STAC Items would depend on GeoJSON.

Hrm, though it looks like you don't actually implement the /service-desc endpoint.

Looking at all the OGC requirements again does make me want to be able to offer something simpler - like just the /search/ endpoint and /conformance with a hardcoded value. Or /collections without more overhead.

Perhaps we have a stac-core that is the landing page (stac catalog and conformance/ (borrowed from / compatible to OAFeat) and collections/ ? With /search optional, and you can use the collectionId's to limit your search, but you don't have to implement collections/{id}/items for all collections - that'd be the stac-items class?

m-mohr:
It's already weekend here, but just quickly some remarks regarding conformance classes: https://earthengine.openeo.org/v1.0 is likely not 100% compliant with Features Core. As you said, service-desc is missing (service-doc as well). It was always a pain that OGC requires them and personally I think they are usually not very useful anyway. But regardless, although I have copy&pasted http://www.opengis.net/spec/ogcapi-features-1/1.0/conf/core into the /conformances, it is not conformant as I have not implemened the .../items/... endpoints, but those are required by core, I think. Its just there because there's nothing else to put there yet. Once we have the STAC conformance classes, I'll likely just list stac-collections and drop the features conformance or replace it with OGC API Commons conformance classes...

cholmes:
> Maybe we can discuss a bit more on Monday during this late call?

You mean the one at 1pm pacific I was going to do with Matt? That'd be great if you could join, though that is quite late. I could also try to get up early, like 6am, and call before Lumi (hopefully) wakes up.

cholmes:
Thanks for the late-night thoughts, it's helpful to me. I'll see if I can come up with something, though may not find much work time between now and monday.

Digging into OGC Common a bit it seems like they don't have the simplicity we're hoping for. Which is a bummer because I got excited about that idea:

They seem to require /api on the very core - http://docs.opengeospatial.org/DRAFTS/19-072.html#resource-requirements-section

And then the collections conformance class seems to talk about items and querying them: http://docs.opengeospatial.org/DRAFTS/20-024.html#rc_collections-section

Why oh why? 

So I think we should just probably decide what our 'core' is, layer on /search, and have optional OAFeat conformance classes.

But yeah, would be good to talk through what our 'core' is - could be more than one conformance class potentially. I do feel it does hit up against the core structure of child vs collection stuff, but maybe we can get a plan for it, or figure out how to get an iterative step out now.

m-mohr:
I don't think the documents you are referring to are up-to-date. Afaik, /api is not required itself, although service-desc/doc are required als links, which must point to something that was previously at /api. It's basically the same, just not at a specific location. Unfortunately, this is also required in Features (and most of us ignore that), so there's no real difference between Features and Commons, I think.
I think they also slimmed down /collections to not talk about items and querying any longer. So I think we could still follow Commons, but likely need to check with them... finding the latest draft was always a bit messy :-|

matthewhanson:
Still trying to wrap my head around this, so based on our conversation Monday there should be a conformance class for STAC root catalog, and the bare minimum STAC "API" would conform to this + GeoJSON conf class...right? Also, technically as we discussed, these two conf classes could apply to a static catalog as well.

So conf classes can't "depend" on others?

cholmes:
My notes from monday are we'd do:

* stac-api core - This is the landing page (/) and 'conformance'. Matthias and I decided in another issue that conformance for us should actually be in the landing page, using the exact same conformance json structure (so it's nicer for clients). We could theoretically have stac core only have conformance at /, but I think it probably makes sense to have it at / and /conformance. The landing page has links to child and/or data (collections), which are discovered by looking at the landing page.

stac-search - Is a conformance class for the search endpoints. Matthias said we could split into post and get. I'm inclined to have the conformance class require GET, with POST optional. It seems like clients could discover on their own if POST was supported.

ogc api commons part 2: collections - Conformance class if you support collections.

ogc api features - for collections/{collectionId}/items and querying that. Would be features core, features geojson and features openapi conformance classes.

stac-spec - If implementing OGC API collections and/or commons this is used to say that the returned collections and items are STAC compliant.

I remember us talking about the geojson requirement class in our 'core', but I think that may not be right - need to understand if OGC API - Features geojson requirements class is truly stand alone and reusable, or if it depends on core. 

As for conf classes 'depending' on others, I'd guess the clearest is to declare them all, so a client can look for what it needs, instead of having to know somehow that the class they're looking for is implicitly declared in another class.

matthewhanson:
Ok, this sounds good to me, an example landing page would be useful inline here, to indicate what links would match up with which conformance classes.

m-mohr:
We could theoretically have stac core only have conformance at /, but I think it probably makes sense to have it at / and /conformance. 

I think it totally makes sense to just have / in core and leave /conformance to OGC, saying that if OGC API - Features is implemented, /conformance should reflect the conformances in /. I don't see any benefit to have it in core.


> stac-search - Is a conformance class for the search endpoints. Matthias said we could split into post and get. I'm inclined to have the conformance class require GET, with POST optional. It seems like clients could discover on their own if POST was supported.

I think I'd split them as they are quite a bit different and it's easier to discover, but it's not a strong preference. I also don't have a better argument yet. ;-)

> I remember us talking about the geojson requirement class in our 'core', but I think that may not be right - need to understand if OGC API - Features geojson requirements class is truly stand alone and reusable, or if it depends on core.

No, it's geojson that is implcitly (or explicitly) defined through stac-search and stac-spec. Core itself has no relation with geojson (the landing page is just a catalog).

> As for conf classes 'depending' on others, I'd guess the clearest is to declare them all

Yeah, I think I agree... but not 100% sure yet. ;-)

Could we quickly come up with a full list of names/URLs here? I think it would be useful to structure the repo and OpenAPI files etc...

cholmes:
> > We could theoretically have stac core only have conformance at /, but I think it probably makes sense to have it at / and /conformance.

> I think it totally makes sense to just have / in core and leave /conformance to OGC, saying that if OGC API - Features is implemented, /conformance should reflect the conformances in /. I don't see any benefit to have it in core.

Cool, I'm convinced.

> I think I'd split them as they are quite a bit different and it's easier to discover, but it's not a strong preference. I also don't have a better argument yet. ;-)

I think I'd prefer to start with less conformance classes then expand.

> No, it's geojson that is implcitly (or explicitly) defined through stac-search and stac-spec. Core itself has no relation with geojson (the landing page is just a catalog).

Ah, right, I forgot that just collections wouldn't have any geojson.

> Could we quickly come up with a full list of names/URLs here? I think it would be useful to structure the repo and OpenAPI files etc...

That'll be my priority for tonight, will try to at least commit a rough draft. Then will circle back from there tomorrow and flesh out the text. Good idea on structuring the openapi files.

cholmes:
Ok, didn't quite get as far as I wanted, but hopefully there's enough to give feedback on - I'm sure I had some things off.

I was most unsure about the dependency on commons. @m-mohr - I couldn't figure out if the commons collection conformance meant you needed to implement /conformance. 

The other thing I'm not sure about is how you would have a server that had OGC API features collections where some were STAC and some were just Features (like buildings - something without assets). The way that made sense to me is that all are valid features, and then some would be STAC, and you could have them all on the same landing page / collection list. But it seems like with our 'stac response' (what I named 'stac spec', since that seemed ambiguous, but I'm open to other names), then that wouldn't be possible? Since you'd declare the stac response conformance class, and then clients would expect that everything conforms to STAC. Unless you listed them from another landing page. Then I wonder if it's really necessary, etc.

Any feedback is welcome, I'm not attached to anything here, just trying to get down what we talked about, and I'm sure some things are off.

cholmes:
@m-mohr + @matthewhanson - ok, made more progress, redid the overview to try to give a good intro to things as we discussed. I'll continue on, but would be great to review that, and the conformance classes at the bottom (I'll move them up somewhere most likely)

api-spec.md
```diff
@@ -6,7 +6,139 @@ The STAC API defines a RESTful JSON-based server to query [SpatioTemporal Asset
...
+off point for the more powerful capabilities - it contains a list of URL's, each described by particular link 
+'relationships' (`rel`) to indicate their functionality. 
+
+The `search` rel is one of the most common, often located at a `/search` endpoint. It re-uses all of the OAFeat [query
+parameters](http://docs.opengeospatial.org/is/17-069r3/17-069r3.html#_items_) specified in their 'core', and adds a 
+couple more. It does not require a full implementation of OAFeat, it is instead a simplified construct that can run
+a search across any set of indexed STAC [`Items`](stac-spec/item-spec/README.md). 
+
+The other common most common link relationship is `data`, which goes to a complete list of available 'Collections', along
+with a mechanism to request individual collections by ID. These are specified by OGC API, in the
```
matthewhanson, line 25: I'm not clear on what this means "along with a mechanism to request individual collections by ID"
Does it mean that by using the `data` rel type you are implementing the /collections endpoint?

api-spec.md
```diff
@@ -6,7 +6,139 @@ The STAC API defines a RESTful JSON-based server to query [SpatioTemporal Asset
...
+a search across any set of indexed STAC [`Items`](stac-spec/item-spec/README.md). 
+
+The other common most common link relationship is `data`, which goes to a complete list of available 'Collections', along
+with a mechanism to request individual collections by ID. These are specified by OGC API, in the 
+'[Collections](http://docs.opengeospatial.org/DRAFTS/20-024.html#rc_collections-section)' section of OGC API Common. 
+STAC's [Collection](stac-spec/collection-spec/README.md) extends the OGC Collection with a handful of [additional 
+fields](stac-spec/collection-spec/collection-spec.md#collection-fields). STAC API's are expected to return STAC
+compliant Collections.
+
+Some STAC API implementations just use it to describe their Collections, without providing search of individual
```
matthewhanson, line 31: "just use "it to describe their collections....not clear what "it" refers to, the `data` rel type?  Collections endpoint?

api-spec.md
```diff
@@ -6,7 +6,139 @@ The STAC API defines a RESTful JSON-based server to query [SpatioTemporal Asset
...
+items. Many implementations go further than just providing the `search` rel by becoming fully compliant with
+[OGC API - Features Core](http://docs.opengeospatial.org/is/17-069r3/17-069r3.html#_requirements_class_core), implementing
+following their [GeoJSON](http://docs.opengeospatial.org/is/17-069r3/17-069r3.html#_requirements_class_geojson) and 
+[OpenAPI](http://docs.opengeospatial.org/is/17-069r3/17-069r3.html#_requirements_class_openapi_3_0) options, as STAC 
+is always in GeoJSON and OpenAPI is used to specify STAC API. Full compliance involves splitting STAC `Items` into
+individual `/collections/{collectionId}/items` endpoints that expose querying single collections, as OAFeat does
+not currently allow cross-collection search. And it adds a few other requirements, explained below. 
+
+Both STAC API and OAFeat then each specify a number of 'extensions' that can be added for additional functionality.
```
matthewhanson, line 40: "then each specify a number of 'extensions'"

replace with 
"allow 'extensions'

api-spec.md
```diff
@@ -6,7 +6,139 @@ The STAC API defines a RESTful JSON-based server to query [SpatioTemporal Asset
...
+}
+```
+
+There are a few things to note: 
+
+- The returned JSON is a valid [STAC Catalog](stac-spec/catalog-spec/catalog-spec.md), and it can provide any number of 'child' links
+to navigate down to additional Catalogs, Collections & Items.
+- The `links` section is a required part of STAC Catalog, and serves as the list of API endpoints. These can live at any location, the 
+client must inspect the the `rel` to understand what capabilities are offered at each location.
+- The `conformsTo` section follows exactly the structure of OGC API - Features [declaration of conformance
```
matthewhanson, line 117: I'm not completely sold on breaking from OGC API, how many users are really going to use the conformance classes enough so they should be at landing page?  Don't the provided links indicate what capabilities are provided?  I'm not sure it's worth doing something different.

api-spec.md
```diff
@@ -287,3 +419,83 @@ All Extensions **should** use attribute names qualified from the root of Item, r
...
+### Commons
+
+[OGC API - Common - Part 2: Geospatial Data](http://docs.opengeospatial.org/DRAFTS/20-024.html) specifies a [Collections requirements 
+class](http://docs.opengeospatial.org/DRAFTS/20-024.html#rc_collections-section) that describes how to request information on Collections
+from an API. This provides the `data` rel type, that links to a list of collections that can all be referred to by ID. It is most often used
+in conjunction with STAC Response, to return valid STAC Collections.
+
+### STAC Response
+
+The STAC Response conformance class specifies that returned OGC Collections should be valid STAC Collections, and that returned OGC Features
```
matthewhanson, line 462: Shouldn't it also specify that returned catalogs are valid STAC catalogs?

matthewhanson (commented):
Overall this looks really good, left a few minor comments and questions.

api-spec.md
cholmes: Yes. I was attempting to make it so things specified the actual endpoints less. But I think OGC spells out the actual endpoint here, so I should probably just say it directly. 

I do think this is an area where we need clarity on 'browse' vs 'ogc'. I see in earth search you use child to go to collections. Which I think makes sense. I would have thought OpenEO would do the same, but they use data and collections/ endpoint. @m-mohr told me a reason that made sense, I think perhaps it's that with a strict STAC thing /collections/ should just be a bunch of child links - it shouldn't inline the actual collection information, like @matthewhanson does in https://earth-search.aws.element84.com/v0/collections

api-spec.md
cholmes: Will clarify, but yes - just use the collections endpoint (linked to with the data rel type) to describe collections, without returning items.

api-spec.md
cholmes: I can still go either way. Was compelled by @m-mohr's arguments, I think the main ones are https://github.com/opengeospatial/oapi_common/issues/71 I tend to think about being nicer to clients, particularly javascript ones, and saving an extra request just to get conformance seems nice. I wouldn't say it's 'breaking' with OGC API, it's just a nice shortcut if OGC API is implemented, and if it's not implemented then it's a straightforward path.

cholmes: Oh, and the point to me isn't users - it's programmatic clients, that read the conformance classes to figure out the capabilities. Like can it make a 'sort' or 'fields' query, or does it have to request everything and do sort / fields on the client side? Is CQL supported? Etc. I think most API clients will want to know what extensions are available, so let them do it with their first request.

api-spec.md
cholmes: I can call that out, but a STAC Collection *is* a STAC Catalog, so that's probably why I left it out...

matthewhanson: In that case it should say "OGC Catalogs" because that would include Catalogs and Collections, but saying "OGC Collections" indicates it doesn't include Catalogs that aren't Collections.

m-mohr: There is no "OGC Catalog" defined AFAIK.

cholmes:
Ok, I think I'm to a decent draft, though still need to do the 'extensions' section, which should be easy.

@m-mohr + @matthewhanson - the one area that didn't work so well was the Collections and STAC Response (what I called the conformance class that says this returns STAC). Collections felt really lame since OGC hasn't released any drafts, so to be sure I got the right version I'd link to https://github.com/opengeospatial/ogcapi-common/blob/cc8ca2f011d7e1b19721268c4bf2b97c163d160a/20-024.pdf or https://github.com/opengeospatial/ogcapi-common/tree/cc8ca2f011d7e1b19721268c4bf2b97c163d160a/collections

I'm strongly leaning towards not declaring a dependency on OGC API Commons Part 2: Collections, but just sub-setting the OGC API Features collection section, specifying it in STAC, and noting that we aim to align with OGC API Common when it is released. 

I'm also thinking about just eliminating the STAC Response / stac-spec thing. Not sure if you saw my comment above:

> The other thing I'm not sure about is how you would have a server that had OGC API features collections where some were STAC and some were just Features (like buildings - something without assets). The way that made sense to me is that all are valid features, and then some would be STAC, and you could have them all on the same landing page / collection list. But it seems like with our 'stac response' (what I named 'stac spec', since that seemed ambiguous, but I'm open to other names), then that wouldn't be possible? Since you'd declare the stac response conformance class, and then clients would expect that everything conforms to STAC. Unless you listed them from another landing page. Then I wonder if it's really necessary, etc.

I think the conformance is all about the API capabilities, and these are really about the content response. Perhaps at some point OGC will have a way to indicate content response compliance. I wonder in the meantime if we could just add a `type` to our links, for the 'stac media type' (which is an open issue https://github.com/radiantearth/stac-spec/issues/918). 

But for now I'm inclined to just leave it off of 'conformance'. Making our own Collection conformance class makes this easier - `search` returns STAC Items, and `collections` return STAC collections. If you additionally declare OAFeat then collections could return non-STAC collections and non-STAC items (in addition to STAC items).

m-mohr:
Just a quick remark before a full review: I think conformance classes can also give hints about the response bodies. GeoJSON is doing that (response body is geojson), CRS extension is doing that (response body includes some fields for CRS), ... Other than that they use OpenAPI for documenting the API and checking for validity, I think.

Regarding mixed collections and items (features and stac): That is something OGC also needs to think about. Not sure they have a final answer, but I think Features has a itemType field in Collections at least. Not 100% sure though it's meant to distinguish between features, coverages, ...

I'll dig deeper tomorrow...

cholmes:
> Just a quick remark before a full review: I think conformance classes can also give hints about the response bodies. GeoJSON is doing that (response body is geojson), CRS extension is doing that (response body includes some fields for CRS), ... Other than that they use OpenAPI for documenting the API and checking for validity, I think.

That's fair, I was probably just looking for excuses to not do the stac response thing, but the main reason is the mixed collection stuff. If you've got a straightforward way to go about it I'm open to it. But I'm really getting into 'don't let great be the enemy of good' mode, and slightly regretting taking on the 'conformance' thing, as I've literally spent the entire day on stac-api release stuff, in addition to a number of hours on my vacation, so want to just get an approach down. I think we can do a beta.2 that iterates to more improvements, and that will also be easier when there is a more stable OGC API base to build upon.

cholmes:
Looked into itemType a bit - I could get behind it. It's referenced more in records api, where they set it to 'records'. I could see us saying 'stac' or 'stac-item' as the itemType. You specify it in the collection response. For OGC it looks like it is just used to indicate what the items are, but I think that's mostly because they are item-centric, so I think we could just use it. 

So then the STAC Response conformance class would just indicate that some of the collections are STAC, indicated by itemType = STAC? I don't know that I see the big win of making that its own conformance class though, seems like we could just indicate in STAC API Core that if you add STAC collections or Items that should be indicated by itemType set to STAC.

I'm still hesitant to rely on OGC collections when it is not a locked-in version. @m-mohr - I also feel like it doesn't really meet your use case, as it has `/items` in the same conformance class. Might be good for you to advocate to them for tighter conformance classes, including one that lets you just return 'collections' to describe a layer, without having to return 'items'.

m-mohr:
Okay, this is now a really large PR with a massive amount of comments. It's hard to follow all discussions and answer them in the limited amount of time I have atm. I'm going through the api-spec quickly to see what comes to my mind and list it in random order:

- We'll need to update the OpenAPI files to reflect the comformances property on the landing page
- From the examples it's not clear how to detect from the landing page whether POST search is supported or not. Either we need to include the `method` in the link or we need to use two conformance classes.
- More for readability in the tables, we can shorten the conformance URIs a bit, I think, for example the core class to <http://stacspec.org/spec/api/1.0.0-beta.1/core>
- I think we should split this into multiple pages, one page per conformance class. Very much like extensions. And then each extension is also a conformance class basically (i.e. there are no extensions any more, just conformance classes).
- The chapter "HTTP Request Methods and Content Types" is a bit out of place. I think we need to move some parts to conformance classes and some parts can be general API guidelines.
- I'm not sure yet about the dependency management of conformance classes, so I asked here: https://gitter.im/opengeospatial/WFS_FES?at=5fc67c7ffeecfc6f390a24bd That will likely guide us a way forward for the OACommon conformance classes etc. itemType could be a solution or a set of conformance classes, let's see.
- I think we can simplify the STAC API table and remove some conformance classes that are "just" dependencies, e.g. GeoJSON and OpenAPI specification. That can then be shown in the respective pages of the conformance class depending on it. The initial table should just show "our" conformance classes and basically be an overview table like in the STAC spec extensions.

@cholmes In general, I think you did an outstanding first draft. I liked quite a lot, now we need to get the details done, but I'm happy to help now that I have a first overview of the PR.

Shall I start to restructure a bit? We should discuss who does what at which time to avoid heavy merge conflicts etc.

cholmes:
@m-mohr - If you could start the restructure that'd be great. How about you take lead for the rest of your day today? I'm busy the next 1.5 hours, so if you could commit at that point I may be able to help for an hour or so after that. And then I can take over at 4pm my time. If you could focus on the openapi stuff that'd be great, and feel free to start the general restructure.

m-mohr:
@cholmes Okay, let's regularly communicate status in our private Gitter conversation.

cholmes:
> From the examples it's not clear how to detect from the landing page whether POST search is supported or not. Either we need to include the method in the link or we need to use two conformance classes.

I'm cool with either approach.

> More for readability in the tables, we can shorten the conformance URIs a bit, I think, for example the core class to http://stacspec.org/spec/api/1.0.0-beta.1/core

+1

> I think we should split this into multiple pages, one page per conformance class. Very much like extensions. And then each extension is also a conformance class basically (i.e. there are no extensions any more, just conformance classes).

Seems reasonable, though I think there's still a place for extensions - maybe it's just less mature components. Like 'version' I'd leave as an extension. Though I do worry about going too far, making the spec annoying to read / get your head around. Though I suppose a nice overview could help - summarize each section a bit and link down.

My other thought was that eventually we should get to OGC style 'requirements'. Like their numbered, testable assertions. I had been thinking that each conformance class would have one of those, but that the api-spec.md would be the 'overview', which would talk about each of them, describe in enough detail and examples to get your head around. That's what I was aiming for with the current prose.

> The chapter "HTTP Request Methods and Content Types" is a bit out of place. I think we need to move some parts to conformance classes and some parts can be general API guidelines.

Agreed. I was going back and forth on removing it entirely. But agreed on restructuring it.

> I'm not sure yet about the dependency management of conformance classes, so I asked here: https://gitter.im/opengeospatial/WFS_FES?at=5fc67c7ffeecfc6f390a24bd That will likely guide us a way forward for the OACommon conformance classes etc. itemType could be a solution or a set of conformance classes, let's see.

Sounds good.

> I think we can simplify the STAC API table and remove some conformance classes that are "just" dependencies, e.g. GeoJSON and OpenAPI specification. That can then be shown in the respective pages of the conformance class depending on it. The initial table should just show "our" conformance classes and basically be an overview table like in the STAC spec extensions.

+1 - I was initially using that table to organize my thinking / keep ogc conformance classes straight / make sure I had everything. But agree that they don't add much value in the current table.

matthewhanson:
I agree with restructuring this to make it more digestable.

I'll be around tonight for reviewing this.

m-mohr:
@matthewhanson This is not ready at all. I'm in the middle of the restructuring, most things are still broken. Will probably take some more days, but of course you can check whether it looks reasonable on the higher level.

m-mohr:
Is there a good way to use OpenAPI for extensions? I'd like to somehow get rid of the buggy merging tool, but can't find an alternative except structuring the extensions as "optional" dependencies (i.e. search would include sort etc)

matthewhanson:
Not sure, but was going to look into that as part of moving the generation of the files into Circle.

m-mohr:
@matthewhanson That is something we should align on. I'm happy for any solution, but I'd need to know how to structure it now. Would it make sense to discuss in-person in the next days?

cholmes:
Ok, just got in a number of commits, looking to clean up and clarify things, and get things into the right format. 

I do really want to try to release this week, and am making this my top priority, so would be really great if the two of you can catch up soon on the OpenAPI approach @matthewhanson and @m-mohr. Feel free to meet early / find a time and add me as optional.

Overall the structure is feeling quite good - thanks for all your work on this @m-mohr. The one thing that does feel weird to me is having just one 'core' and then everything in 'extension'. Item-Search & Collections feel 'more core' to me, soon followed by Features. And then 'version' and 'transaction' feel a bit 'more extension-y' than the rest of the extensions. And version even more so since afaik there aren't implementations, it's just a proposal.

So I think it'd be good to regroup things a bit, in terms of folder structure. Perhaps core, item-search and collections are in one group, then features on its own, and 'extensions'. I could also see core, then have item-search folder also include sort, fields, context, query (to be replaced by filter). Those have always felt fairly 'core' to me, for a rich search service, but it's nice to have them as options. And then collections and features live in their own directories (I could perhaps see an argument to share an 'OGC' directory when we have collections as OACommon). And 'extensions' would have transactions and version.  Or an alternate would be 'core', 'extensions' and 'proposals', and move 'version' to proposals. Not really attached to any particular scheme, but I do think the folder structure should help communicate to new implementors a bit. Having a super small core and then a bunch of different options all bundled in 'extensions' doesn't help them figure out what to do next after 'core'. Yes, this can be mitigated by a nice overview, but I don't like relying on people for sure reading that - I often will just start clicking around a repo and exploring (and yes, OGC repo's frustrate me).

I did also flesh out 'collections' more, to be an independent STAC thing, based on OAFeat's descriptions, and noted that we based it on those, and are following Common Part 2 Collections and will align to that when it is solid.

The next things on my list are more overview & fixing links. Oh, @m-mohr - what's your plan for a combined documents? There's broken links to STAC.yaml, STAC-Extension.yaml, OAFeat.yaml, etc. I guess OAFeat I can direct to the extension folder. For now I'm leaving the broken ones in so we're sure to update them.

m-mohr:
Thanks for doing all the fixes, looks good. I don't think I agree with commit f691ca292133d1b8f603ad61f632e17f7345280a though. That is text, which is mostly specific to item-search and doesn't fully makes sense in the overview IMHO.

I'm not sure yet how to best group, may idea was to have sort etc be re-usable also for collection search for example.

STAC.yaml etc. should be hosed somewhere else and be generated automatically. Will discuss with Matt, maybe we can host the at the conformance URIs.

m-mohr:
I've restructured as discussed with Matt, there's now fragments and extensions. Don't have the time right now to explain it in detail, let's discuss later. I've continued to fix OpenAPI files, I still need to work on Transactions and Version at least.

m-mohr:
Okay, I managed to restructure/fix all the OpenAPI files.

@matthewhanson You can start with the CI work.

m-mohr:
This is now the chance to add an ItemCollection fragment in a place where it really fits, see https://github.com/radiantearth/stac-api-spec/issues/25

cholmes:
+1 on getting ItemCollection right.

I redid f691ca2 and then cleaned it up so it only talks about search. It was weird to me it talked about other stuff, so that's why I put it in overview, where it's cool to touch on everything. So just made it so it makes sense in this file. May do some sort of http section in the overview.

m-mohr:
@matthewhanson @cholmes I've moved folders and fixed the OpenAPI refs, now there are likely some broken links.

Some thoughts:
- core, item-search, ogcapi-... on the same level as fragments feels weird
- There are no openAPI files for item-search-sort, item-search-query etc. They are all merged into the openapi.yaml for item-search. Implications?
- Do we need the extensions folder in ogcapi-features or can we just place version and transaction one level up?
- The item-search extensions still need to be moved from the Readme

cholmes:
Awesome, thanks!

> core, item-search, ogcapi-... on the same level as fragments feels weird

I get it, but can't yet think of any other good alternatives. Could put core,item-search,etc in 'spec' or something, but then not sure that helps.

> There are no openAPI files for item-search-sort, item-search-query etc. They are all merged into the openapi.yaml for item-search. Implications?

So search and query live in fragments? But then the item search openapi.yaml includes those fragments? It seems like an ideal could be an openapi.yaml that's just the core, and openapi-ext.yaml that includes them? But if that's not possible I think it's ok.

> Do we need the extensions folder in ogcapi-features or can we just place version and transaction one level up?

One level up like ogcapi-features/transactions and ogcapi-features/version instead of ogcapi-features/extension/transaction? 

I'd lean towards ogcapi-features/extensions/ that has transactions.md & transactions.yaml in it. Not set on any of it though.

> The item-search extensions still need to be moved from the Readme 

I can do that.

@m-mohr or @matthewhanson - can one of you approve so I'm able to merge? I'll fix the links to get CI building, and do some clean up before I merge. But want to be able to do it today and have us shift to PR's that are easier to review, as discussed.

matthewhanson (approved):

Issue #55, opened 2020-11-12
lock versions of STAC core and Features API


cholmes:
We should have released versions of STAC API mark certain versions of underlying specs as dependencies, and be clear on what version they depend on up front. 

We should update all the links to be sure they track the actual version. This will likely involve find & replace for each release, unless we figure out some way to automatically do it.

 - [x] Update submodule to point at STAC core 1.0.0-beta.2
 - [x] Add note about properly checking out submodules
 - [x] Update stac-spec links to be relative
 - [x] Make clear the features API dependency
 - [x] Add what updates are needed for a 'release' (update submodule, stac links)

m-mohr:
Agreed for Features API, but couldn't/shouldn't the API be able to operate on multiple STAC versions? Also, if we fix it to STAC, we also need to re-release once a new STAC version is out. Maybe just specify a version range (e.g. >= 0.9.0 and < 2.0)?

cholmes:
Ah, that's fair. 

I was mostly thinking about making sure the link references to the STAC spec used the 'tags', instead of just master (which may change and possibly break the links).

Should we say 0.9.0 and above for STAC API beta.1? I guess your < 2.0 is just signaling that STAC API doesn't guarantee to work with a breaking change?

cholmes:
Actually I'm realizing that the 'stac-spec @ 568a048' external folder thing in the repo is maybe meant to help with links not breaking? I was starting to do full URL references to STAC spec. But it looks like I can do relative ones to that folder, which we can 'upgrade' at our leisure? 

Seems like we should probably 'set' it to 1.0.0-beta.1 of STAC spec for now? And then I can update the links to refer to it relatively. I have no idea how to work with those. But if that route makes sense I can dig in - I'm guessing that's why we had it in there in the first place?

m-mohr:
Yeah, makes sense to set it to beta.2 and make relative links.

cholmes:
Ok, I'll change them to relative links. 

Though do we need to explicitly tell people to get the sub-modules? I just have empty directories, and reading https://stackoverflow.com/questions/3796927/how-to-git-clone-including-submodules it looks like that is the norm. Should we remind people how to check it out to be sure to get the right stac-spec so the relative links work?

m-mohr:
Hmm... yeah submodules are tricky. I guess most people read it online, but a note somewhere doesn't hurt, right?

m-mohr:
[...]

PR #65, opened 2020-11-18
Updates for relative links and being clear on version dependencies

cholmes:
**Related Issue(s):** #55 

**Proposed Changes:**

1. Add note about properly checking out submodules
2. Update stac-spec links to be relative
3. Make clear the features API dependency
4. Added a PROCESS.md to have a place to put reminders for release.
5. Add what updates are needed for a 'release' (update submodule). Ended up making it so you don't need to update
the stac-spec version - just instructed people to check the CHANGELOG of the submodule to see.

**PR Checklist:**

- [ ] OpenAPI files only: I have run `npm run generate-all` to update the [generated OpenAPI files](https://github.com/radiantearth/stac-api-spec/blob/dev/README.md#openapi-definitions).

### DIFF
--- a/PROCESS.md
+++ b/PROCESS.md
@@ -1,0 +1,72 @@
+## STAC Development & Release Process
+
+### Development Process
+
+The SpatioTemporal Asset Catalog API specification is under active development. 
+
+The `master` branch aims to always be stable, meaning that all the pieces of the specification are consistent and well
+explained, and all the examples are consistent with the specification. The `dev` branch is a place of active development, 
+where a new change in one part of the spec might not yet be fully updated everywhere else. The team uses the 
+[stac-api-spec issue tracker](https://github.com/radiantearth/stac-api-spec/issues) to identify and track all that will be done for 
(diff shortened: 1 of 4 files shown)


PROCESS.md
```diff
@@ -0,0 +1,72 @@
+## STAC Development & Release Process
```
m-mohr, line 1: I'm wondering whether we should have a central place for things that are common across stac-spec and stac-api-spec. It's always a pain to keep things aligned...

m-mohr (approved):
Seems like a good start. Maybe add a CHANGELOG entry for the new definition that this API version only supports spec 0.9 up to 2 (exclusive)?

PROCESS.md
cholmes: I'm open to it, but not sure how it would work. I feel like it either is harder for a new user to understand (go look at this repo to find out x, y & z), or it's some complicated technical thing (though maybe submodules come close)?

I sorta figured the release processes would also diverge some.

PR #72, opened 2020-12-04
Fragments

cholmes:
**Related Issue(s):** #

**Proposed Changes:**

1. Flesh out fragments in various ways
2. Changed language in extensions for the links to their fragments to 'defined by' instead of 'dependency', as the latter made it seem like it was something more. I did keep the fragment 'dependents' the same though, since there could be other extensions that actually extend the fragment potentially. 
3. Added the item-search extensions, in the main item-search readme. Moved the examples to their own document as it got to be a long document with those in there, and we use examples in other files most other places.

**PR Checklist:**

- [ ] OpenAPI files only: I have run `npm run generate-all` to update the [generated OpenAPI files](https://github.com/radiantearth/stac-api-spec/blob/dev/README.md#openapi-definitions).


cholmes:
I may do some more work on this, but marking as ready for review as it's a chunk of work that improves things. I'm thinking of just holding off on trying to do the ogcapi-features extensions for beta.1, since we didn't really do them before.

I did just put the item-search extensions all in the item-search document, but I do think they should evolve to be each in their own file or possibly even folder. I do think we should get to the testable assertions style that OGC has (though those should be the last thing people read, instead of the main spec narrative), and then it'd make sense to have each of those in an extension.

fragments/README.md
```diff
@@ -1 +1,17 @@
 # STAC API - Fragments
+
+This directory contains reusable OpenAPI fragments that are used by other capabilities. They tend to 
+be parameters that get mixed in to various endpoints, and keeping them in this separate fragment directory
```
m-mohr, line 4:
```suggestion
be common API components such as parameters that get mixed in to various endpoints.
They are kept in this separate fragment directory as they are not specific to any extension and are meant to be re-used when drafting new API extensions.
```

fragments/fields/README.md
```diff
@@ -1,19 +1,15 @@
-# STAC API - Fields Fragment
+  # STAC API - Fields Fragment
```
m-mohr, line 1:
```suggestion
# STAC API - Fields Fragment
```

fragments/fields/README.md
```diff
@@ -1,19 +1,15 @@
...
 
 - **OpenAPI specification:** [openapi.yaml](openapi.yaml)
 - **Fragment [Maturity Classification](../../extensions.md#extension-maturity):** Pilot
 - **Dependents:**
   - [Item Search](../../item-search)
```
m-mohr, line 7: Maybe start here with a introduction what the fragment is meant to achieve. It starts with too many details (e.g. content-types).

fragments/query/README.md
```diff
@@ -2,16 +2,15 @@
...
   - [Item Search](../../item-search)
 
-The STAC search endpoint, `/search`, by default only accepts a limited set of core filter parameters.
-The Query API extension adds additional filters for searching on the properties of Items.
+The `query` parameter adds additional filters for searching on the properties of Items.
 
 The syntax for the `query` filter is:
 
-```js
+```json
```
m-mohr, line 13: Recommend to revert this change as the example is invalid JSON and JS renders this a bit better (i.e. I chose JS specifically)

item-search/README.md
```diff
@@ -75,6 +49,8 @@ GET /search?collections=landsat8,sentinel&bbox=-10.415,36.066,3.779,44.213&limit
 }
 ```
 
+For more examples see [examples.md](examples.md).
```
m-mohr, line 52: Maybe also put the link into the "intro" list, here people may miss it.

item-search/README.md
```diff
@@ -180,124 +156,75 @@ searching on specific Item properties.
...
+field names to sort by, with an indication of direction. It can be used with both GET and POST, the former using '+' and
+'-' to indicate sort order, and the latter including a 'direction' field in JSON. The full description of the semantics
+of this extension can be found in the [sort fragment](../fragments/sort).
 
-Request:
-```http
-POST /search
-Search-After: LC81530752019135LGN00
-```
+STAC API's that support the query functionality must include the conformance class
```
m-mohr, line 212: The conformsTo details are repeated many times. Maybe there a more clever way without writing so much? The shorter the docs, the more likely people read it. Is it enough to specofy Conformance URI at the top and have somewhere centrally defined that you need conformance classes? At a place you can't miss it. A bold note in the core or so...

m-mohr (approved):
Good work. I added some comments and suggestions, but nothing major.

fragments/query/README.md
cholmes: Ah, gotcha - didn't realize it was deliberate.

item-search/README.md
cholmes: what's the 'intro' list?

m-mohr:
We also need to discuss extensions and Conformance URIs a bit more, e.g. currently there's a # in the item-search extensions as the don't have a separate OpenAPI and folder. It's different for ogcapi-features extensions, we should think about a guideline maybe how to structure new extensions in the future.

item-search/README.md
cholmes: Ah, that's fair. I think for these I was going more towards the 'requirements' style, to make clear exactly what you have to do. But my plan is to do actual requirements in a next release, in their own document, and to keep this one as the narrative. So I think it's ok to just remove some of the detail/repetition.

m-mohr: The list with maturity, conformance uri etc.

item-search/README.md
m-mohr: The list with maturity, conformance uri etc.

fragments/fields/README.md
cholmes: Will do.

item-search/README.md
m-mohr: We still need to document conformsTo on the landing page anyway, we have not done that yet: https://github.com/radiantearth/stac-api-spec/issues/27#issuecomment-738883830

cholmes:
Sounds good on discussing on monday. I do think it'll probably make sense to put the item-search extensions in folders eventually, with readme, examples and 'requirements' - but they won't have an openapi yaml, since it's supplied by the fragment. 

I addressed most of the comments, so will merge this in. The one I didn't understand was on the examples:

> Maybe also put the link into the "intro" list, here people may miss it.

Not sure which intro list you're talking about.

item-search/README.md
m-mohr: line 3-6 in the file

m-mohr:
> Not sure which intro list you're talking about.

@cholmes line 3 - 6 in the file. The introductory list that has the important details, could as last item has a link to examples.

cholmes:
Got it. That still took me a minute - I kept interpreting it as add a link at that line to the list at the top. Adding a link to examples directly from the top makes way more sense.

m-mohr:
Oh, yeah, sorry, sometimes one is on the wrong track and then it's hard to get them back on the right track ;-) Can be merged!

PR #86, opened 2020-12-08
OpenAPI + CI

matthewhanson:
This is the 'dev' branch because I was using it to test the deployment of the `dev` versions of the OpenAPI files.  Updates to dev will be immediately deployed to the `dev/` folder in gh-pages

As in stac-spec, when a tag is created these files will be published to `vXX` folder.

**Related Issue(s):** #
#12 

**Proposed Changes:**

- bundle each openapi.yaml doc (resolving fragments)
- publish yaml to gh-pages
- create combined version of STAC + all extensions


m-mohr:
Could we migrate the bash script to node to be platform independent? I can't run this on Windows...

matthewhanson:
@m-mohr I spent several hours trying to get a Node script working to do the build using speccy and in the end got sick of Promises and wrote that bash script in a few minutes.  Just wanted to get it working.

Let's get this merged in so we have it building and then we can replace it with Node.

For Redoc we can just reference the published URLs from stacsite.  I'm still figuring out the combine, there's a couple errors in the yaml that swagger-combine doesn't like, working on getting those resolved, but we can do a different PR for that.

cholmes (approved):

m-mohr:
I can do the Node script for beta.2 @matthewhanson

Issue #111, opened 2021-03-16
Change root endpoint (/) the be a Catalog, rather than a Collection or a Catalog


philvarner:
Right now, the text spec just says that it's JSON, but the OpenAPI definition says it can be either a Catalog or a Collection.  I think it would be cleaner if this was only a Catalog. Allowing it to be either makes it a bit more complex to interact with, and I feel that generally an API will be built over a Catalog rather than a single Collection.

cholmes:
+1 - makes sense to me. Would be weird if it's just a collection, and if an implementor wants to do just have one collection they can just put a catalog that links to one collection.

matthewhanson:
Agree as well, Catalog requirement makes sense.

philvarner:
I'm going to add this to the description of Catalog in stac-spec and put a PR in to update this repo.

PR #112, opened 2021-03-17
Tidy up catalog references

philvarner:
**Related Issue(s):** #111

**Proposed Changes:**

1. make explicit that / returns a Catalog and not a Collection (this is implied by the schema, but not explicit in the textual description)
2. add `type` field to Catalog openapi definition and example

**PR Checklist:**


m-mohr:
As you've added the type field for Catalog, I'd think it should also add the type field for Collections, right?

Issue #114, opened 2021-03-22
Broken stac-spec links


duckontheweb:
Links to entities defined in the [radiantearth/stac-spec](https://github.com/radiantearth/stac-spec) repo are set up as relative links within the `stac-spec` folder. Since this is actually a submodule these links don't work as expected and should maybe be updated to be absolute URLs.

I can work on a PR to fix this if it's not already underway.

cholmes:
So the problem with absolute links is keeping them up to date and properly pinned to the right STAC release. We don't want to just go to master since that may change. The submodule seemed to give us control and easy updates. 

Is it not working even when you've checked the sub-module out correctly? Or are you just saying that people shouldn't have to check out the sub-module to have links work right?

duckontheweb:
The problem I'm seeing is with the README hosted on GitHub. Someone going to https://github.com/radiantearth/stac-api-spec and clicking on the link to one of those objects (e.g. the `Catalog` link in the "About" section of the main README) gets a 404 response.

cholmes:
Fair point, and yeah, that seems less than ideal. I wonder if they used to work, as it seems like github has changed some of their links around recently. 

A PR to fix would be great, and then perhaps also file a ticket to spur us to get a robust way to update the links when we release.

duckontheweb:
Actually, using a [`git subtree`](https://www.atlassian.com/git/tutorials/git-subtree) instead of a `git submodule` might solve our problem here. Subtrees actually copy the child repo into the parent, which would keep links intact (versus submodules, which just point to a commit on the child repo). It requires `git>=v1.7`, but that version is years old, so it shouldn't be a problem.

m-mohr:
@duckontheweb I think I'd be in favor of subtree than absolute links, if we can make it that PRs to the subtree get rejected or so.

cholmes:
+1 on the subtree. To me that's ok even if we can't get automatic rejection of the PR's, since we'll soon turn on the requirement of 2 reviewers as we have in stac-spec. Reviewers would know that edits to stac-spec aren't supposed to be done here.

duckontheweb:
> @duckontheweb did your PR resolve this issue?

Yes, you can close this. I'll also close #115, which is now no longer necessary.

PR #115, opened 2021-03-26
Relative -> absolute links for stac-spec references

duckontheweb:
**Related Issue(s):** #
#114 

**Proposed Changes:**

1. Use absolute instead of relative links for all references to pages in the `radiantearth/stac-spec` repo. 
All URLs are pinned to `v1.0.0-rc.1`.

**PR Checklist:**

- [x] ~I have added my changes to the [CHANGELOG](https://github.com/radiantearth/stac-api-spec/blob/dev/CHANGELOG.md) **or**~ a CHANGELOG entry is not required.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,16 +8,19 @@ The SpatioTemporal Asset Catalog (STAC) specification aims to standardize the wa
 A 'spatiotemporal asset' is any file that represents information about the earth captured in a certain space and 
 time. The core STAC specification lives at [gitub.com/radiantearth/stac-spec](https://github.com/radiantearth/stac-spec).
 
-A STAC API is the dynamic version of a SpatioTemporal Asset Catalog. It returns a STAC [Catalog](stac-spec/catalog-spec/catalog-spec.md), 
-[Collection](stac-spec/collection-spec/collection-spec.md), [Item](stac-spec/item-spec/item-spec.md), 
+A STAC API is the dynamic version of a SpatioTemporal Asset Catalog. It returns a STAC 
+[Catalog](https://github.com/radiantearth/stac-spec/blob/v1.0.0-rc.1/catalog-spec/catalog-spec.md), 
+[Collection](https://github.com/radiantearth/stac-spec/blob/v1.0.0-rc.1/collection-spec/collection-spec.md), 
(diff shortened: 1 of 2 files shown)


matthewhanson:
@duckontheweb we've been linking to the master branch of the stac-spec from the extensions, rather than to the v1.0.0-rc.1 tag.   It saves us from having to update the links (which once we get to 1.0.0 should be an uncommon occurrence), but there's also a risk that links could become broken.

For now, I think linking to master branch is probably best.   thoughts @cholmes ?

cholmes:
I'm definitely of two minds on this. I do think the rc.1 link is less good, because it's going to be rc.2 soon and then 1.0.0. 

But I don't love going straight to master, since if stac spec were to go to 2.0 and then someone was looking at STAC API release 1.0.0 they'd follow a link there, which isn't really what STAC API is based on. 

I think best may be to do 'master' for right now, but once STAC 1.0.0 is at then we could change it to link there...

duckontheweb:
Yeah, I'm a little back-and-forth on this, too. It's not hard to update the `stac-spec` version in these links, so pinning to a specific version seemed like an easy way to make sure they don't break. On the other hand, this version of the API spec doesn't really "depend" on the content of `v1.0.0-rc.1` of `stac-spec` in any specific way that I can see, it only depends on the fact that files stay in the same place within that repo, so maybe pointing to `master` is fine and we just know that if we change the structure of that repo then we need to update these links as well.

duckontheweb:
@matthewhanson @cholmes I updated the links to point to `master` instead of `1.0.0-rc.1`

m-mohr:
Why not just update the submodule?

duckontheweb:
@m-mohr See #114 for the reasoning behind this. Relative links to a submodule do not work correctly on GitHub since the directory doesn't actually exist in that location.

cholmes:
@duckontheweb - are you able to update this PR so it uses a sub-tree? Or would that need to be a new PR? Or some other way to get that working?

duckontheweb:
I opened #119 and marked it as WIP until we confirmed that's the way we wanted to go.

PR #119, opened 2021-04-20
[WIP] Core as subtree

duckontheweb:
**Related Issue(s):** 
#114 

**Proposed Changes:**

1. Make `stac-spec` a `git subtree` instead of a `git submodule`

**PR Checklist:**

- [x] ~I have added my changes to the [CHANGELOG](https://github.com/radiantearth/stac-api-spec/blob/dev/CHANGELOG.md) **or**~ a CHANGELOG entry is not required.


duckontheweb:
@cholmes @matthewhanson @m-mohr This a demo of what it would look like have `stac-spec` embedded as a subtree instead of a submodule. A couple of things to note:
* This is currently using `v1.0.0-rc.2` of `stac-spec`. When we want to update, we will need to use `git subtree pull` from the new core spec version.
* Links to core spec docs (e.g. Catalog, Collection) will be to files *in this repo* and not in radiantearth/stac-spec.
* I added an item to the PR checklist to help contributors remember not to change anything in the `stac-spec` directory. If a subtree seems like the right way to go, I'll look into adding a CI check as well.

duckontheweb:
Also, looks like the failing tests are due to invalid links to the `stac-spec/extensions` folder that no longer exists now that extensions have been moved.

duckontheweb:
I'll fix the broken extension links and add a CI check so you all can review this properly.

duckontheweb:
@cholmes @m-mohr @matthewhanson This is ready for review. 

I added an `npm` command to check for changes to the `stac-spec` directory against the given revision/branch. It will return a -1 exit code if there have been changes. Also added this as a job in the CircleCI config.

package.json
```diff
@@ -15,7 +15,8 @@
     "check-openapi-item-search": "spectral lint item-search/openapi.yaml --ruleset .circleci/.spectral.yml",
     "check-openapi-fragments": "spectral lint fragments/*/openapi.yaml --ruleset .circleci/.spectral-fragments.yml",
     "build-openapi": ".circleci/build-openapi.sh",
-    "publish-openapi": "node .circleci/publish.js"
+    "publish-openapi": "node .circleci/publish.js",
+    "check-stac-spec-changes": "git diff --quiet HEAD ${npm_config_compare_to} -- stac-spec"
```
m-mohr, line 19: This is meant to be used by CI only, right? As I think this would fail on any other system, maybe just use that command directly in the CI without exposing it as npm script?

m-mohr (approved):
Not ideal, but probably the best solution that we can get right now. Thanks for doing it. Left one comment, otherwise lgtm.

PR #128, opened 2021-05-10
Filter Extension

philvarner:
**Related Issue(s):** #32

**Proposed Changes:**

1. Defines Filter Extension for allowing CQL queries.

Non-changes
- @cholmes 's original PR removed `intersects`. I think that's a separate concern and should be in a different PR (and needs a lot of feedback and discussion)

Open Issues / Questions
- The OpenAPI spec for it still has some validation problems, but it gets the correct ideas across.
- How do the conformance classes relate to implementing filter on both /collections/{collectionId}/items and /search? Are both required?

**PR Checklist:**


philvarner:
Open Questions:
- how do the conformance classes relate to implementing filter on both /collections/{collectionId}/items and /search? Are both required?

jisantuc:
I've been working on an implementation in Franklin using only CQL JSON and here are my comments so far:

### Strange mix of types?

One thing that I keep tripping on is a pretty strange mixture of types. An expression for filtering a field is a _single key_ at the root of an object, followed by one of a map also holding a single key, or an array of maps holding a single key. This singularity strikes me as odd, because objects can hold many keys, so if there's only supposed to be one thing, why does the data structure in principle hold more than one? Similarly, for `or` and `and`, the value behind the key is supposed to be, e.g.:

```json
[
  {"eq": ...},
  {"neq": ...}
]
```

These string keys in a by-default plural structure are much less convenient to work with than an alternative in which they live under a `kind` key in each json object, like:

```json
[
  {"kind": "eq", ...},
  {"kind": "neq", ...}
]
```

I think we're getting this from the OGC spec itself, so there might not be much we can do about it, but it's pretty weird to program with. I've been fighting a JSON round trip test for about 90 minutes while tabbing back and forth to the spec to try to get this to cooperate, and a better matching of types to what they represent would have made this part much easier.

### `queryables` seem pretty tough to know ahead of time

This is becoming a recurring theme with the open-endedness of the spec, but it's not obvious to me what I'm supposed to do for `queryables` in a server implementation when the server will serve other people's data. It's also not clear to me whether `queryables` is required. I can accumulate item properties as someone fills Franklin's database with items (the dynamic case mentioned in the spec), but then it's not obvious to me what "queryable" means -- does it mean it's technically possible to query based on `x`, or that querying based on `x` will be fast? If the latter, then I need to expose some way for users to control what's queryable and what isn't, since Postgres JSONB indexing is really effective for equality at a key filters but pretty much worthless otherwise. There's a related issue for this in Franklin from before the CQL work here https://github.com/azavea/franklin/issues/616 -- basically there's a trade-off in what fields we can search efficiently and index size / cost of inserting items, and it's hard to make a good choice about that without knowing what promise "x is queryable" represents.

It's also pretty difficult in the dynamic case to figure out what to do with fields that don't have a JSON schema. Let's say someone posts an item that has a string value in `mySpecialField` -- if I'm dynamically aggregating queryables, I don't know that all `myVerySpecialField` values are strings, just that I happen to have seen one.

### Overall though pretty solid 👍🏻

This isn't really that different technically from the previous query extension in its implementation for us -- build up the cases that something of the type `CQLFilter` can have, destructure with a `fold` or a match to whatever the `CQLFilter` has to become. A bunch of stuff was copy-pastable from the query extension implementation we had in Franklin.

### Random notes

I appreciated the clarification that "POST with a JSON body to the Items resource is not supported, as POST is used by the Transactions Extension for creating Item objects." Since I'm working 100% on the JSON side of things that was going to come up for me eventually, and when I got to thinking about it I saw it was already there. 💯

It wasn't too difficult I think to implement the RHS property name lookup Postgres. One thing that wasn't clear to me was whether the properties in any part of the queries were _only_ top-level properties or could drill into the JSON object. I acted as if they could access nested fields, but only if it's strictly through a hierarchy of objects. Something I think would be a cool capability would be having some kind of JSON / textual representation of object traversal (in the [traversal systems](https://chrispenner.ca/posts/traversal-systems) / optics sense) that would allow quite precise querying, e.g. "the name of the seventh band in eo:bands is aerosol", but that's way out of scope here 🤷🏻‍♂️

m-mohr:
@jisantuc This is great feedback, but it seems most should be actually be pointed towards OGC API - Features so that they may improve things depending on your feedback. :-)

jisantuc:
I was asked for comments on the PR via email so I put them here first

philvarner:
**Queryables**

The spec uses the words that the queryables "may" and "can" be used in filter expressions, which I interpret to mean that using a parameter name that's not in queryables should result in an error. Maybe one recommendation would be to allow wildcards, like `properties.*`, as queryables, so a user could query for any value in Properties? Or maybe there's a wildcard option `*` for queryables that indicates that the filter is allowed to use property names not defined explicitly as queryables, but there's no guarantee they'll work (e.g., if I query `foobar = 1` and foobar doesn't exist, I get a successful 0 results).

philvarner:
> One thing that wasn't clear to me was whether the properties in any part of the queries were only top-level properties or could drill into the JSON object. 

You could do this with queryables -- a queryable `foo` that actually maps to `properties.bar.baz.foo`

> I acted as if they could access nested fields, but only if it's strictly through a hierarchy of objects. Something I think would be a cool capability would be having some kind of JSON / textual representation of object traversal (in the traversal systems / optics sense) that would allow quite precise querying, e.g. "the name of the seventh band in eo:bands is aerosol", but that's way out of scope here 🤷🏻‍♂️

Yes, I think we have many use cases like this, particularly eo:bands, as users search multiple products for assets that have the spectral bands they need. Might be something GraphQL could express well.

fragments/filter/README.md
```diff
@@ -0,0 +1,632 @@
...
+
+(TBD: there is a proposal to allow finding what queryables are available for a subset of collections, e.g., `/queryables?collections=c1,c3`).  
+
+The Landing Page endpoint (`/`) will have a Link with rel `http://www.opengis.net/def/rel/ogc/1.0/queryables` with an href to the endpoint `/queryables` (TBD: hopefully, this will have a parameter `collections` defined on it to query for the intersection of the queryables in a group of collections). Additionally, each Collection resource will have a Link to the queryables resource for that collection, e.g., `/collections/collection1/queryables`. 
+
+The queryables endpoint returns a JSON Schema describing the names and types variables that may be used in filter expressions. This response is defined in JSON Schema because it is a well-specified typed schema, but it is not used for validating a JSON document derived from it. This schema defines the variables that may be used in a CQL filter.
```
cholmes, line 145: Might be nice to explain the 'simple' path to doing this. Like the next sections explain all these complex possibilities, that make it seem a bit scary to implement this whole extension. I think it'd be good to be clear about the core, minimum that a STAC implementor needs to do in order to get Filter extension to work. Do they need to make every field queryable? Is there a minimum implementation that they could almost hard code that returns datetime, geometry, other required fields? Could they hardcode the common metadata that 'might' appear? Or easily code up something that checks common metadata and includes it, but isn't trying to construct JSON Schemas on the fly from arbitrary STAC data?

I want to think about how we make an easy onramp for a STAC implementor, so they can do the minimum and have it be valuable. Perhaps put some stuff in an 'advanced' section.

fragments/filter/README.md
```diff
@@ -0,0 +1,632 @@
...
+- xtraplatform-spatial has a CQL [ANTLR 4 grammer](https://github.com/interactive-instruments/xtraplatform-spatial/tree/master/xtraplatform-cql/src/main/antlr/de/ii/xtraplatform/cql/infra)
+
+## Queryables
+
+The Queryables mechanism allows a client to discover what variable terms are available for use when writing filter
+expressions.  These variables can be defined per-collection, and the intersection of these variables over all collections is what is available for filtering when there are no collection restrictions. These queryables are the only variables that may be used in filter expressions, and if any variable is used in expression that is not defined as a queryable, a 400 Bad Request exception should be returned.
+
+Implementers should add queryables for all root Item fields (e.g., id, collection, geometry) with those names. Fields in Item Properties should also be exposed with their names, and not require expressions to prefix them with  `properties`. **Everything else should be fully-qualified? How to do more deeply-nested queries and those on lists is an open question. How to disambiguate names that could appear in multiple places in an Item?**
```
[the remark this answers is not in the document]
cholmes: I now see there's examples below. Might be good to have a short one up here, and then link down to fuller ones below.

fragments/filter/README.md
```diff
@@ -0,0 +1,632 @@
...
+    },
+    "geometry" : {
+      "description" : "Geometry",
+      "$ref": "https://schemas.stacspec.org/v1.0.0/item-spec/json-schema/item.json#/geometry"
+    },
+    "datetime" : {
+      "description" : "Datetime",
+      "$ref": "https://schemas.stacspec.org/v1.0.0/item-spec/json-schema/datetime.json#/properties/datetime"
+    },
+    "eo:cloud_cover" : {
```
cholmes, line 235: This is one thing that's not clear to me, which may be worth explaining - if it's listed in queryables does that mean it's in all the data? Or can you put things in queryables that might be there? Like is the contract of it that the data will have this value? Or just that this is the name of the field to be filtered on, without a guarantee that it's there. If the latter then it could be interesting to just define the STAC API root queryables as the common metadata plus stable fields, and that API implementations can just put that up, without having to do any dynamic stuff.

philvarner:
**Strange mix of types?**

`cql-text` is pretty nailed down, since they're just formalizing something that's been used in practice for a while. But CQL JSON is new, and I think they're pretty open to suggestions still, particularly if they're related to implementation difficulties. I didn't look at it too closely because it looked good enough, and no one is going to actually be manually constructing these queries anyway. Also, I feel Scala exposes these type issues a lot more than other common languages, since there's just more correct automatic behavior if you can get the types structured right.

fragments/filter/README.md
```diff
@@ -0,0 +1,632 @@
...
+
+Note: the GET examples with query parameters are unescaped to make them easier to read. 
+
+### Example 1
+
+This example uses the queryables definition in (Interaction with Endpoints)(#interaction-with-endpoints).
+
+#### Example 1: GET with cql-text
+
+Note that `filter-lang` defaults to `cql-text` in this case, so this is only shown for completeness.
```
cholmes, line 273: Is this doing a filter-crs default like in the POST one? Perhaps should mention it?

fragments/filter/README.md
```diff
@@ -0,0 +1,632 @@
...
+
+Note that `filter-lang` defaults to `cql-text` in this case, so this is only shown for completeness. 
+
+```http
+GET /search?filter-lang=cql-text&filter=id='LC08_L1TP_060247_20180905_20180912_01_T1_L1TP' AND collection='landsat8_l1tp'
+```
+
+#### Example 1: POST with cql-json
+
+Note that `filter-lang` defaults to `cql-json` and `filter-crs` defaults to `http://www.opengis.net/def/crs/OGC/1.3/CRS84` in this case.
```
cholmes, line 281: If they are defaulting then why are they included in the example? 

I like to show the first examples with as few parameters as possible, and note the defaults that came into use with it, right after the example.

fragments/filter/README.md
```diff
@@ -0,0 +1,632 @@
...
+        "landsat8_l1tp"
+      ]
+    ]
+  }
+}
+```
+
+### Example 2
+
+This example uses the queryables definition in (Interaction with Endpoints)(#interaction-with-endpoints).
```
cholmes, line 305: Should perhaps note that the collection field only really makes sense for item-search queries, not for ogcapi-features queries, as those are already against a single collection.

fragments/filter/README.md
```diff
@@ -0,0 +1,632 @@
...
+      "description" : "Sentinel Data Coverage",
+      "type": "integer",
+      "minimum": 0,
+      "maximum": 100
+    }
+  }
+}
+```
+
+Note that `sentinel:data_coverage` is a proprietary extension (e.g., not defined in a formal, public way), and must be defined directly within the schema.
```
cholmes, line 449: I think the wording here could use tweaking. I don't think 'proprietary; means 'not defined in a formal, public way'. I think we can just call it an 'unspecified extension' or something? I do think we also want to have data providers define schemas, even if they aren't in a public, stac schemas extension. Like ideally there's a sentinel:data_coverage schema that sentinel defines. I know we're not there yet, but I think it could be good to make it seem like more of a special case to not have a schema. Like say that when STAC extensions have not yet defined their fields in a JSON Schema they can directly define it in queryables, but maybe make it a more 'custom' example than sentinel.

fragments/filter/README.md
```diff
@@ -0,0 +1,632 @@
...
+                        [-77.0824, 38.7886], [-77.0189, 38.7886],
+                        [-77.0189, 38.8351], [-77.0824, 38.8351],
+                        [-77.0824, 38.7886]
+                    ]]
+                }
+        ]
+    },        
+}
+```
+
```
cholmes, line 623: Not required for getting the PR in, but it'd be nice to have some examples of the more advanced features, to show how powerful it can be. Maybe it's in a separate document, so we don't overwhelm people with a super long page of examples.

cholmes (commented):
Looking good! Just put up a bunch of review comments. I think the biggest ones are:

* I think we should have two conformance classes - one for item-search, and one for ogcapi-features. The latter we just use OGC's class directly, and the other we define our own class (which I think you already did). But let people choose one or the other or both. Some people implement 'item-search' without ogcapi-features, so we should give them the option to just do the item-search CQL. I think this approach should simplify a few bits, as the fragment can talk through the common stuff, and then do more specifics in each section (like cross collection search in item search). I noted in comments that that was the intent behind the fragments, to define conformance classes in multiple places, but we never executed on it. fields and sort should do that next - like use them in ogcapi-features under stac conformance namespace (until we manage to fully align)

* Clarity on what to do with the fact the spec may change with more conformance classes. This is mentioned a good bit, but I think we should try to make the recommendation of what to do right now clear. I think it's to use their conformance classes but it's ok to not fully implement everything, and in a future version there will be the right conformance classes for more of a subset.

* Queryables - it reads now like there's potentially a lot of work to do here. Should focus on what the minimal requirements are, and make it clear for the STAC content. I think the examples helped make it a bit more clear to me, but we should just structure for an easy on ramp for new implementors, where they can just follow some simple recommendations and be 'good enough'.

cholmes:
I saw @jisantuc's review come in as I was most of the way through mine, so there's some overlapping stuff, but I didn't go back and tweak them all. It seems like he had some overlapping thoughts/questions on queryables, and Phil answered some of them:


> The spec uses the words that the queryables "may" and "can" be used in filter expressions, which I interpret to mean that using a parameter name that's not in queryables should result in an error. Maybe one recommendation would be to allow wildcards, like properties.*, as queryables, so a user could query for any value in Properties? Or maybe there's a wildcard option * for queryables that indicates that the filter is allowed to use property names not defined explicitly as queryables, but there's no guarantee they'll work (e.g., if I query foobar = 1 and foobar doesn't exist, I get a successful 0 results).

Something in those ideas to add flexibility would make sense to me, but not quite sure how either of those would work - may need to see some examples. But thinking about the easy 'on-ramp' for a STAC developer, it'd be useful if we could just provide them with almost a 'static' queryables document that they could just use. At the very least it could define datetime, geometry and collection. It seems like it'd be nice if we could also define common metadata and stable extensions as 'queryables' in some static way. But I'm not sure how that plays with OGC if the data doesn't actually have that field.

fragments/filter/README.md
```diff
@@ -0,0 +1,632 @@
...
+## Limitations of Item Search 
+
+OAFeat defines a limited set of filtering capabilities. Filtering can only be done over a single collection and 
+with only a single `bbox` (rectangular spatial filter) parameter and a single datetime (instant or interval) parameter. 
+
+The STAC Item Search specification extends the functionality of OAFeat in a few key ways:
+- It allows cross-collection filtering, whereas OAFeat only allows filtering within a single collection. (`collections` parameter, accepting 0 or more collections)
+- It allows filtering by Item ID (`ids` parameter)
+- It allows filtering based on a single GeoJSON Geometry, rather than only a bbox (`intersects` parameter)
```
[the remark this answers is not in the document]
philvarner: >  But the intersects parameter is definitely simpler, as you don't need to include the name of the geometry and figure out as much of how CQL works. 

Yes, this is where I stand. I think people should be about to implement intersects without even needing to touch queryables and CQL.

philvarner:
@cholmes @jisantuc I just filed this issue with OGC https://github.com/opengeospatial/ogcapi-features/issues/582

On the one hand, I like the tightness of having queryables defined and allowing that type of dynamic discovery, but on the other hand, it severely complicates and/or restricts filtering when the Items have arbitrary values in their properties. I think as OGC has defined it, it's too tight to work in many situations, and there needs to be a way for the client and/or server to determine the semantics of what they allow/deny.

fragments/filter/README.md
```diff
@@ -0,0 +1,632 @@
...
+- Simple CQL (`http://www.opengis.net/spec/ogcapi-features-3/1.0/req/simple-cql`) - defines the query language used for the `filter` parameter defined by Filter
+
+This STAC Filter extension extends the Filter conformance class such that these parameters also apply
+to the STAC Item Search resource (/search). The OAFeat Filter conformance class already requires that these
+parameters work for GET requests to the Items resource (/collections/collectionId/items). POST with a JSON body to the Items resource is not supported, as POST is used by the Transactions Extension for creating Item objects.
+
+Additional, the implementation must support at least one of "CQL Text" (`http://www.opengis.net/spec/ogcapi-features-3/1.0/req/cql-text`) or "CQL JSON" (`http://www.opengis.net/spec/ogcapi-features-3/1.0/req/cql-json`). If both are advertised as being supported, it is only required that both be supported for GET query parameters, and that only that CQL JSON be supported for POST JSON requests. It is recommended that clients use CQL Text in GET requests and CQL JSON in POST requests.
```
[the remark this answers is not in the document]
philvarner: > (more line breaks would be good here to comment more precisely)

I copied the remark linting config from stac-spec the other day, and apparently the 150 line limit is commended out in that config? Anyway, that's why remark wasn't telling me there was a problem, so I'm going to re-enable it and fix all these long lines.

fragments/filter/README.md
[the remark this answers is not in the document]
cholmes: Ok. I do think if that's our stance (which I'm cool with) we should at least try a bit to advocate to OGC to make it an extension for OGC API - Features. So instead of the 'intersects CQL class' we'd have an OGC parameter that is 'intersects'. For OGC API - Features they'd want that as an extension, but for STAC we'd make it a requirement.

Feel free to give that feedback, or I'm also happy to do so.

fragments/filter/README.md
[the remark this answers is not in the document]
m-mohr: You must have looked at a very old version. Best is probably to simply copy/paste https://github.com/radiantearth/stac-spec/blob/master/.circleci/rc.yaml

fragments/filter/README.md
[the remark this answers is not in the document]
philvarner: I'll put that on my todo list.

fragments/filter/README.md
```diff
@@ -0,0 +1,632 @@
...
+multiple spatial or temporal filters.
+
+## Filter expressiveness 
+
+This extension expands the capabilities of Item Search and the OAFeat Items resource with 
+[OAFeat Part 3 CQL](https://portal.ogc.org/files/96288) 
+by providing an expressive query language to construct more complex filter predicates. The operators are similar to 
+those provided by SQL. The Simple CQL conformance class requires the logical operators `AND`, `OR`, and `NOT`; 
+the comparison operators '=', `<`, `<=`, `>`, `>=`, `LIKE`, `IS NULL`, `BETWEEN`, and `IN`; the spatial operator 
+`INTERSECTS`; and the temporal operator `ANYINTERACTS`.
```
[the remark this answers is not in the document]
philvarner: I've rewritten this section, and I think is much clearer now.

fragments/filter/README.md
```diff
@@ -0,0 +1,632 @@
...
+
+The Filter extension **requires** support of these three conformance classes:
+
+- Filter (`http://www.opengis.net/spec/ogcapi-features-3/1.0/req/filter`) - defines the Queryables mechanism and 
+  parameters `filter-lang`, filter-crs`, and `filter`
+- Features Filter (`http://www.opengis.net/spec/ogcapi-features-3/1.0/req/features-filter`) - defines that the parameters defined in `Filter` apply to the Features endpoint (`/collections/{collectionId}/items`) defined by OAFeat Part 1. 
+- Simple CQL (`http://www.opengis.net/spec/ogcapi-features-3/1.0/req/simple-cql`) - defines the query language used for the `filter` parameter defined by Filter
+
+This STAC Filter extension extends the Filter conformance class such that these parameters also apply
+to the STAC Item Search resource (/search). The OAFeat Filter conformance class already requires that these
```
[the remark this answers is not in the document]
philvarner: I realize now that I wrote this up wrong. Filter and Simple CQL are required, Features Filter is recommended, and Filter Extension is what binds Filter and Simple CQL to Item Search, in the same way Features Filter does to the Features endpoint /collections/{cid}/items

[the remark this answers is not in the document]
philvarner: But, I get what you're saying about the conformance classes and how STAC extends them.  I'm going to need to think about this some more.

fragments/filter/README.md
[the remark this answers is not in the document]
philvarner: I must have something wrong with my local dev branch. I'm going to recopy it in.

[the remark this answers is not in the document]
philvarner: > I was also thinking we might make more specific recommendations. Like in item-search filter we require cql-text, in line with our requirement for GET. Also I'm not sure how cql-json is supposed to work for GET? You include JSON in the body? I think we should recommend against that...

cql-json works with GET the same as cql-text does -- the JSON is just a string value to the query parameter.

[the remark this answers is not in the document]
philvarner: This also indicates we may need our own, more specific conformance classes, since we should presumably allow an implementation to only support CQL Text against the OGC Features endpoint, and only support CQL JSON against the /search endpoint, but right now, there's no way to advertise it as that restricted. I don't think anyone will want to do that, but it's an edge case that reveals how fine-grained the conformance classes need to be.

fragments/filter/README.md
```diff
@@ -0,0 +1,632 @@
...
+classes and their operations is both allowed and encouraged. Implementation of these is often limited by the 
+operations supported by the implementation's datastore, for example, Elasticsearch does not support the spatial 
+operations required by the Enhanced Spatial Operators.
+
+There will likely be a change to Simple CQL where this conformance class only requires support of expressions with a property name of the left hand side and a literal on the right hand side (e.g., `eo:cloud_cover <= 10`), and additional conformance classes will support arbitrary uses of properties and literals in expression. The primary motivation for this is to allow implementations that use datastores that do not easily support arbitrary expressions like these to implement Simple CQL (e.g., Elasticsearch). 
+
+There will also likely be a change where the Simple CQL conformance class is decomposed into several other 
+conformance classes to aid composition. After these changes, it is possible that this extension will not require 
+the implementation of the operators IN, BETWEEN, LIKE, and IS NULL predicates.
```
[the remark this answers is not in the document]
philvarner: added

fragments/filter/README.md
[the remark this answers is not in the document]
philvarner: I've been thinking about his since James brought it up, and I think there has to be some way to define different semantics for queryables. I updated the text with a recommendation to allow this if you want, and assume that there will be some way to formally allow it in the future.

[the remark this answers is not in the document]
philvarner: I could also see the queryables semantics changing so that those are the only fields that are guaranteed to be successful in filter expressions, and anything else is open-ended like most NoSQL DB query semantics-- maybe you'll get an error, maybe you just won't get any results because the field doesn't exist or doesn't have the right schema type in the db.

fragments/filter/README.md
```diff
@@ -0,0 +1,632 @@
...
+
+## Queryables
+
+The Queryables mechanism allows a client to discover what variable terms are available for use when writing filter
+expressions.  These variables can be defined per-collection, and the intersection of these variables over all collections is what is available for filtering when there are no collection restrictions. These queryables are the only variables that may be used in filter expressions, and if any variable is used in expression that is not defined as a queryable, a 400 Bad Request exception should be returned.
+
+Implementers should add queryables for all root Item fields (e.g., id, collection, geometry) with those names. Fields in Item Properties should also be exposed with their names, and not require expressions to prefix them with  `properties`. **Everything else should be fully-qualified? How to do more deeply-nested queries and those on lists is an open question. How to disambiguate names that could appear in multiple places in an Item?** 
+
+(TBD: there is a proposal to allow finding what queryables are available for a subset of collections, e.g., `/queryables?collections=c1,c3`).
```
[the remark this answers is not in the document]
philvarner: added

fragments/filter/README.md
```diff
@@ -0,0 +1,632 @@
...
+
+Implementers should add queryables for all root Item fields (e.g., id, collection, geometry) with those names. Fields in Item Properties should also be exposed with their names, and not require expressions to prefix them with  `properties`. **Everything else should be fully-qualified? How to do more deeply-nested queries and those on lists is an open question. How to disambiguate names that could appear in multiple places in an Item?** 
+
+(TBD: there is a proposal to allow finding what queryables are available for a subset of collections, e.g., `/queryables?collections=c1,c3`).  
+
+The Landing Page endpoint (`/`) will have a Link with rel `http://www.opengis.net/def/rel/ogc/1.0/queryables` with an href to the endpoint `/queryables` (TBD: hopefully, this will have a parameter `collections` defined on it to query for the intersection of the queryables in a group of collections). Additionally, each Collection resource will have a Link to the queryables resource for that collection, e.g., `/collections/collection1/queryables`.
```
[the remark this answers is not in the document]
philvarner: I removed that line here, since that parameter doesn't actually exist, and I don't think it's going to be that important.

fragments/filter/README.md
philvarner: I'll do that.

fragments/filter/README.md
philvarner: added text to make this clear.

fragments/filter/README.md
```diff
@@ -0,0 +1,720 @@
...
+}
+```
+
+## Implementations
+
+- [GeoPython PyCQL](https://github.com/geopython/pycql/tree/master/pycql), and the 
+  [Bitner fork](https://github.com/bitner/pycql) to be used in stac-fastapi
+- [Franklin](https://github.com/azavea/franklin) is working on it.
+
+Note that the [xbib CQL library (JVM)](https://github.com/xbib/cql) is the OASIS Contextual Query Language, not
```
m-mohr, line 717: Interesting that Josh actually implemented against another "CQL" library that is not actually OGC CQL... Maybe we should somewhere at the top actually say that there are other "CQL"s out there and that OASIS CQL is different than OGC CQL and people should be careful with that they use etc.

philvarner: I made the same mistake in EOD and also used xbib.  There is a section about this at the top.

m-mohr: Must have missed it, sorry.

extensions.md
```diff
@@ -67,11 +68,12 @@ the service supports. This are listed at the top of each extension description,
 | Extension Name                                                                    | Conformance URI                                                            |
 |-----------------------------------------------------------------------------------|----------------------------------------------------------------------------|
 | [Fields](item-search/README.md#fields)                                            | <https://api.stacspec.org/v1.0.0-beta.1/item-search#fields>  |
-| [Query](item-search/README.md#query)                                              | <https://api.stacspec.org/v1.0.0-beta.1/item-search#query>   |
+| [Filter](item-search/README.md#filter)                                            | <https://api.stacspec.org/v1.0.0-beta.1/item-search#filter>   |
```
philvarner, line 71: this needs to be updated

fragments/filter/README.md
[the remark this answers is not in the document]
philvarner: I thought about this more, and we need our own conformance classes that replicate those defined in Part 3. The primary reason is that the Features resource and /search would have to support exactly the same Part 3 classes if we reuse them, and that may not be the same. This also frees us up to change the semantics if we with, so we could, for example, say "our Filter conformance class allows use of properties not defined in Queryables".

I also change the format of these conformance classes to be /extension/filter rather than /item-search#filter -- I feel the former looks better and reduces coupling to to the other conformance classes.

fragments/filter/README.md
philvarner: https://github.com/radiantearth/stac-api-spec/issues/139

fragments/filter/README.md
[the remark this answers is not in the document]
philvarner: On further though, I changed them back to be like /item-search#filter style.

fragments/filter/README.md
philvarner: tweaked

fragments/filter/README.md
```diff
@@ -0,0 +1,786 @@
...
+[OGC Catalogue Services 3.0 - General Model](http://docs.opengeospatial.org/is/12-168r6/12-168r6.html#62) 
+(including a BNF grammar in Annex B). The CQL JSON format is newly-defined, but also not
+expected to change before final.
+
+It should be noted that the "CQL" referred to here is OGC CQL. It is **not** referencing or related two other "CQL" languages, 
+the [SRU (Search/Retrieve via URL) Contextual Query Language](https://www.loc.gov/standards/sru/cql/index.html) (formerly 
+known as Common Query Language) or the [Cassandra Query Language](https://cassandra.apache.org/doc/latest/cql/) used by the 
+Cassandra database.
+
+## Limitations of Item Search
```
lossyrob, line 75: This section is more a justification for the extension, and perhaps doesn't need to be so prominent. It was a little confusing to me reading through the spec as I was expecting this section to begin talking about what this extensions supports. I'd recommend making a section lower down dedicated to talking about other parts of the spec in comparison or justifying this extension, and have the beginning of this document focused on specifying what the filter extension is and does.

fragments/filter/README.md
```diff
@@ -0,0 +1,786 @@
...
+  - [cql.bnf](https://github.com/opengeospatial/ogcapi-features/blob/master/extensions/cql/standard/schema/cql.bnf)
+  - [cql.json](https://github.com/opengeospatial/ogcapi-features/blob/master/extensions/cql/standard/schema/cql.json)
+  - [cql.yml](https://github.com/opengeospatial/ogcapi-features/blob/master/extensions/cql/standard/schema/cql.yml)
+- A JSON Schema for only the parts of the CQL JSON encoding required by this extension is [here](cql.json)
+- A OpenAPI specification for only the parts of the CQL JSON encoding required by this extension is [here](cql.yml)
+- xtraplatform-spatial has a CQL [ANTLR 4 grammer](https://github.com/interactive-instruments/xtraplatform-spatial/tree/master/xtraplatform-cql/src/main/antlr/de/ii/xtraplatform/cql/infra)
+
+These projects have or are developing CQL support:
+
+- [GeoPython PyCQL](https://github.com/geopython/pycql/tree/master/pycql), and the
```
lossyrob, line 195: Should https://github.com/geopython/pygeofilter be used instead, as it seams to be more aligned with the current OAFeatures Filtering spec? This is a bit of a guess, as I haven't grokked both libraries completely, but this seems to be under more active development

fragments/filter/README.md
```diff
@@ -0,0 +1,786 @@
...
+    },
+    "assets_bands" : {
+      "description" : "Asset eo:bands common names",
+      "$ref": "https://stac-extensions.github.io/eo/v1.0.0/schema.json#/properties/eo:bands/common_name"    
+    }
+  }
+}
+```
+
+Alternately, the client could retrieve the queryables for a single collection with
```
lossyrob, line 398: This wasn't clear to me from the Queryables section; consider mentioning this above. E.g.  "Queryables are advertised via a JSON Schema document retrieved from the /queryables endpoint." -> "Queryables are advertised via a JSON Schema document retrieved from the /queryables endpoint, either from the root to retrieve for the queryables across all collections, or off of a collection route (e.g. /collections/collection1/queryables) for the queryables pertaining to a single collection"

Is it required to implement all of the root queryables endpoint + collection-specific endpoints?

fragments/filter/README.md
```diff
@@ -0,0 +1,786 @@
...
+Alternately, the client could retrieve the queryables for a single collection with 
+`/collections/collections1/queryables`, which may have more queryables than available for the entire catalog, since
+there may be queryables that are only relevant to that collection. 
+
+Notice in this schema that instead of directly defining the type information about each field, we have instead used the JSON 
+Schema `$ref` mechanism to refer to a STAC schema definition. This not only allows the reuse of these JSON Schema definitions,
+but also binds an arbitrarily-named Queryable to a specific STAC field. For example, in the above we know that the 
+`eo:cloud_cover` field is referring to the field of the same name in the EO Extension not because they happen to have the same 
+name, but rather because the `$ref` indicates it. The field could just as well be named "cloud_cover", "CloudCover", or "cc", 
+and we would still know it filtered on the EO extension `eo:cloud_cover` field.
```
lossyrob, line 407: I believe this is the case, but wasn't clear to me on the first readthrough - in a CQL query (either text or JSON), the arbitrarily named Queryable is used in the query rather then a reference to the fully qualified property. E.g. in the case it was called "CloudCover", you'd say "CloudCover <- 20.0". Is this correct?

fragments/filter/README.md
```diff
@@ -0,0 +1,786 @@
...
+## Examples
+
+Note: the GET examples with query parameters are unescaped to make them easier to read. 
+
+The parameter `filter-crs` always defaults to `http://www.opengis.net/def/crs/OGC/1.3/CRS84` for a STAC API, so is not shown 
+in any of these examples.
+
+### Example 1
+
+This example uses the queryables definition in (Interaction with Endpoints)(#interaction-with-endpoints).
```
lossyrob, line 423:
```suggestion
This example uses the queryables definition in [Interaction with Endpoints](#interaction-with-endpoints).
```

fragments/filter/README.md
```diff
@@ -0,0 +1,786 @@
...
+```http
+GET /search?filter=id='LC08_L1TP_060247_20180905_20180912_01_T1_L1TP' AND collection='landsat8_l1tp'
+```
+
+#### Example 1: POST with cql-json
+
+Note that `filter-lang` defaults to `cql-json` in this case. The parameter `filter-crs` defaults 
+to `http://www.opengis.net/def/crs/OGC/1.3/CRS84` for a STAC API.
+
+```http
```
lossyrob, line 439: Perhaps use javascript instead of http, which would render like:

```javascript
POST /search
{ 
  "filter": {
    "and": [
      "eq": [
        { "property": "id" },
        "LC08_L1TP_060247_20180905_20180912_01_T1_L1TP"
      ],
      "eq": [
        { "property": "collection" },
        "landsat8_l1tp"
      ]
    ]
  }
}
```

lossyrob (commented):
Really great stuff - represents a lot of detailed work, really appreciate that you're sorting through this.

One high level comment is that you may want to consider splitting up the README into either separate Markdown documents, pulling some of the justification/point-in-time-specific information into the PR description, and/or ongoing issues or TBDs into GitHub Issues. This might make reading easier for users who want to grok what this extension is quickly, as well as make this document more maintainable as things continue to evolve

philvarner:
> Really great stuff - represents a lot of detailed work, really appreciate that you're sorting through this.
> 
> One high level comment is that you may want to consider splitting up the README into either separate Markdown documents, pulling some of the justification/point-in-time-specific information into the PR description, and/or ongoing issues or TBDs into GitHub Issues. This might make reading easier for users who want to grok what this extension is quickly, as well as make this document more maintainable as things continue to evolve

I agree, it's so large with all the examples that it's a lot to process at once, and breaking it up would be good.

item-search/README.md
```diff
@@ -305,3 +306,15 @@ of this extension can be found in the [sort fragment](../fragments/sort).
 This extension is intended to augment the core ItemCollection responses from the `search` API endpoint with a
 JSON object called `context` that includes the number of items `matched`, `returned` and the `limit` requested.
 The full description and examples of this are found in the [context fragment](../fragments/context).
+
+### Query
```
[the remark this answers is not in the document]
philvarner: I think there's general consensus that it will be deprecated. I'm not so sure about removing it completely, at least until the CQL support is well-supported on all of our common language/DB combinations.

fragments/filter/README.md
```diff
@@ -0,0 +1,786 @@
...
+## Filter expressiveness 
+
+This extension expands the capabilities of Item Search and the OAFeat Items resource with 
+[OAFeat Part 3 CQL](https://portal.ogc.org/files/96288) 
+by providing an expressive query language to construct more complex filter predicates. The operators are similar to 
+those provided by SQL. The Simple CQL conformance class requires the logical operators `AND`, `OR`, and `NOT`; 
+the comparison operators `=`, `<`, `<=`, `>`, `>=`, `LIKE`, `IS NULL`, `BETWEEN`, and `IN`; the spatial operator 
+`INTERSECTS`; and the temporal operator `ANYINTERACTS`. 
+
+The ANYINTERACTS operator has effectively the same semantics as the existing `datetime` filter.
```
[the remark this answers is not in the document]
philvarner: I clarified this.

fragments/filter/README.md
philvarner: I'll think about that. I like having the "motivation" section first, to adequately set the stage for where we're starting with this extension.

fragments/filter/README.md
philvarner: Thanks for the pointer to that. I added in pygeofilter, but I also don't know what either provides. Looks like they do have a cql-json impl, whereas I think pyCQL is only doing cql-text right now.

fragments/filter/README.md
philvarner: Thanks, I clarified.

fragments/filter/README.md
philvarner: Yes, and I've clarified it.

fragments/filter/README.md
philvarner: fixed.

fragments/filter/README.md
philvarner: thanks for the suggestion, done.

philvarner:
Filed these for follow-up work:

* https://github.com/radiantearth/stac-api-spec/issues/144
* https://github.com/radiantearth/stac-api-spec/issues/143

m-mohr:
Unfortunately, I don't have the time right now to review it so leave it up to others. Will just do issues after the merge if something comes up...

cholmes (approved):

Issue #113, opened 2021-03-18
Make item search extensions more prominent


m-mohr:
Recently, I was quite often asked whether the search endpoints aren't a bit limited. I was confused but later found out people just don't always see the extensions at the end of the page. We need to mention them more prominent, because people seem to miss them since we reorganized the repo recently. People are especially asking about the query extension.

cholmes:
That seems reasonable. For the next release we hopefully bring in OGC Filter - maybe we put that in its own 'top level' folder. Or we could do like 'advanced querying' and include context, sort, fields. Hrm, maybe we could just 'rebrand' fragments to that, as I think we agreed we're taking 'itemcollection' out.

m-mohr:
It could be as easy as having a TOC at the top of the pages or something similar,

cholmes:
Ah, good point.

PR #130, opened 2021-05-11
Readme item search reorg

philvarner:
**Related Issue(s):** #113

**Proposed Changes:**

1. add ToC to all pages
2. add text about how Item Search is limited and most behavior will be defined in extensions

**PR Checklist:**

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,6 +3,13 @@
 
 # STAC API
 
+- [STAC API](#stac-api)
+  - [About](#about)
+  - [Stability Note](#stability-note)
+  - [Communication](#communication)
+  - [In this repository](#in-this-repository)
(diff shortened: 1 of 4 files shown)


cholmes:
Looks like this is made to master, not dev? I think with beta.1 release we now have dev vs master for API? It looks like there's more changes in this than I would expect.

philvarner:
> Looks like this is made to master, not dev? I think with beta.1 release we now have dev vs master for API? It looks like there's more changes in this than I would expect.

I did a few in a row, so of course I messed up this one. Fixed, now against dev.

cholmes (approved):

cholmes:
Cool. That's a much more reasonable number of files to review :)

Issue #99, opened 2021-01-14
Add recommendation to enable CORS


cholmes:
CORS helps STAC work well, API's should enable it. See https://github.com/radiantearth/stac-spec/issues/888

philvarner:
We should recommend setting (at least) on the responses:

```
access-control-allow-origin: *
access-control-allow-methods: OPTIONS, POST, GET
access-control-allow-headers: Content-Type
```

and if Transactions is supported, also `access-control-allow-methods` Transactions PUT, PATCH, DELETE

PR #131, opened 2021-05-12
Add CORS recommendation

philvarner:
**Related Issue(s):** #99

**Proposed Changes:**

1. Add recommendation to enable CORS for public APIs

**PR Checklist:**

### DIFF
--- a/implementation.md
+++ b/implementation.md
@@ -4,8 +4,9 @@ This document describes implementation recommendations for a STAC API.
 
 ## CORS
 
-It is recommended that public APIs advertise a permissive CORS configuration so UIs running on a different domain
-may more easily access them.
+Web browsers enforce a mechanism called [Cross-Origin Resource Sharing (CORS)](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS) to prevent 
+malicious access to APIs. It is recommended that public STAC APIs advertise a permissive [CORS](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS) 
+configuration so browser-based UIs running on a different domain may more easily access them.
 
 APIs should acknowledge pre-flight request headers. In general, these header values should be set on responses:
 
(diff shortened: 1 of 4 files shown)


implementation.md
```diff
@@ -0,0 +1,22 @@
+# Implementation Recommendations
+
+This document describes implementation recommendations for a STAC API.
+
+## CORS
+
+It is recommended that public APIs advertise a permissive CORS configuration so UIs running on a different domain
```
m-mohr, line 7: Maybe add a short sentence about what CORS is useful for (JS in Browser)? UIs in general is a bit misleading, i.e. it would not apply to a Java Desktop UI. Or link to a CORS article (e.g. wikipedia).

implementation.md
```diff
@@ -0,0 +1,22 @@
...
+may more easily access them.
+
+APIs should acknowledge pre-flight request headers. In general, these header values should be set on responses:
+
+```
+access-control-allow-origin: *
+access-control-allow-methods: OPTIONS, POST, GET
+access-control-allow-headers: Content-Type
+access-control-allow-credentials: false
```
m-mohr, line 16: I think this (credentials) can be removed. It doesn't need to be set to false because that's the default and we should only set what is required by STAC API itself.

implementation.md
```diff
@@ -0,0 +1,22 @@
...
+
+It is recommended that public APIs advertise a permissive CORS configuration so UIs running on a different domain
+may more easily access them.
+
+APIs should acknowledge pre-flight request headers. In general, these header values should be set on responses:
+
+```
+access-control-allow-origin: *
+access-control-allow-methods: OPTIONS, POST, GET
+access-control-allow-headers: Content-Type
```
m-mohr, line 15: The Transaction extension uses the `If-Match` header, which we should mention...

m-mohr (commented):
Some comments to resolve, otherwise LGTM

cholmes (approved):
Looks good, I agree with Matthias's comments.

implementation.md
philvarner: I made this more clear.

implementation.md
philvarner: Added a reference to that.

implementation.md
philvarner: modified. I didn't realize that actually "The only valid value for this header is true (case-sensitive). If you don't need credentials, omit this header entirely (rather than setting its value to false)."

philvarner:
@m-mohr ready for re-review if you want to

m-mohr (approved):

Issue #105, opened 2021-02-27
Use consistent capitalization, links to STAC specs


cholmes:
In https://github.com/radiantearth/stac-spec/pull/1018 we did a consistency pass, would be good to do the same for the API spec at some point.

* Capitalize Item, Collection & Catalog if refering to the main JSON constructs. 'catalog' lowercase can be used to refer to an overall implementation, 'Collection' can refer to the JSON of a collection, or the set of Items within a Collection.
* Link to the core specs at the first occurrence in each file.

PR #132, opened 2021-05-12
link and make more consistent uses of Catalog, Item, and Collection

philvarner:
**Related Issue(s):** #105

**Proposed Changes:**

1. cleanup text to make uses of Item, Collection, and Catalog more consistent. 

**PR Checklist:**


CHANGELOG.md
```diff
@@ -54,5 +54,5 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 See the [stac-spec CHANGELOG](https://github.com/radiantearth/stac-spec/blob/v0.9.0/CHANGELOG.md)
 for STAC API releases prior to or equal to version 0.9.0.
 
-[Unreleased]: <https://github.com/radiantearth/stac-api-spec/compare/master...dev>
```
m-mohr, line 57: Should still be upper-case as the heading above is also upper-case (first letter)

.circleci/rc.yaml
```diff
@@ -40,4 +41,13 @@ plugins:
     - space 
 # Tables
   - remark-lint-table-pipes
-  - remark-lint-no-literal-urls
```
m-mohr, line 43: Could you give some details on these changes? The remark rules on all STAC repos are aligned and to give predicatable behavior I'd suggest to keep them as before...

ogcapi-features/openapi.yaml
```diff
@@ -31,7 +31,7 @@ paths:
             (link relations `service-desc` and `service-doc`),
             the Conformance declaration (path `/conformance`,
             link relation `conformance`), and the Feature
-            Collections (path `/collections`, link relation
```
m-mohr, line 34: Maybe also check whether the original in ogcapi-features has this issue and fix it there, too... Makes alignment in the future easier for us...

ogcapi-features/openapi.yaml
```diff
@@ -97,7 +97,7 @@ paths:
         - Features
       summary: the feature collections in the dataset
       description: |-
-        A body of Feature Collections that belong or are used together with additional links.
```
m-mohr, line 100: I'd rather stick to the original. In OGC API Features they don't use the appended "object" to entities.

m-mohr (commented):
Overall LGTM. Added a couple of comments regarding alignment with STAC/OGC.

.circleci/rc.yaml
philvarner: This one was apparently out of sync with the one in stac-spec, which is why it was giving so many errors. I updated it with that one and the rules still work fine.

CHANGELOG.md
philvarner: this was a remark warning, reverted now.

ogcapi-features/openapi.yaml
philvarner: reverted this and the other one in ogcapi

.circleci/rc.yaml
m-mohr: Ah, maybe we did mess up with dev/master here? Not sure, I thought I had updated all repos already.

m-mohr (approved):

Issue #116, opened 2021-03-27
Recommend html landing page? 


cholmes:
Just came across https://franklin.nasa-hsi.azavea.com/ and it's got a really nice html landing page. 

And reminded me that I really like that OGC API - Features often have html landing pages. But when you give people a root stac URL it's usually a bunch of JSON. 

I'd ideally like to make some 'strongly recommended' thing to have people put an html page at their root catalog as the default, with a link (rel alternate) to the json (which would rel alternate html back). This could just be a stac browser (though we'd need to figure out how it can do rel alternate to the json if it doesn't already), or it could be an overview page like franklin.

philvarner:
One challenge here is that it would prevent a user from easily looking at the JSON responses in the browser. I do this a lot. 

It seems that Franklin no longer has this page, maybe for that reason? @jisantuc 

One thing we seem to be missing in the the Core text docs (but is in the openapi.yaml) is that a user should implement the rel service-docs, which can be nice landing page. This example https://eod-catalog-svc-prod.astraea.earth/api.html is redoc with a little customization (logo, front text, etc).

cholmes:
I don't think it'd prevent easily looking at the JSON responses. Indeed I'd put in a recommendation to have a clear link to it. I like how pygeoapi does it https://demo.pygeoapi.io/master

![pygeo](https://user-images.githubusercontent.com/407017/118181117-38b56d80-b3ec-11eb-90ee-310e48f2fb9b.gif)

Emphasizing rel service-docs a bit more does make sense to me as well, though I feel like a landing page is more links to the collections, overview of the data that is there.

philvarner:
That seems to use an additional query parameter "f=json", which we'd have to define. I'd don't necessarily like that, since then there's Accept and `f` to process.

jisantuc:
> One challenge here is that it would prevent a user from easily looking at the JSON responses in the browser. I do this a lot.

That's part of why we removed it. I told people a few times about Chrome and Firefox extensions that let them set the Accept header, and that was a nuisance, so it's gone.

[...]

PR #133, opened 2021-05-14
add service-doc to core links table, add links diagram

philvarner:
**Related Issue(s):** #116

**Proposed Changes:**

1. add service-doc to Links page
2. add Links diagram

**PR Checklist:**


cholmes (approved):
Looks good. I think my only question with this is if we want to make service-desc and service doc to be 'optional' elements in the core spec? 

The reasoning would be that today if you wanted to include those right now you're supposed to fully implement the features API. Of course nothing would stop you from including them, so maybe it doesn't really matter. I think it'd just be a bit more endorsement to recommend people use it even if they're just implementing core + item-search.

philvarner:
> Looks good. I think my only question with this is if we want to make service-desc and service doc to be 'optional' elements in the core spec?
> 
> The reasoning would be that today if you wanted to include those right now you're supposed to fully implement the features API. Of course nothing would stop you from including them, so maybe it doesn't really matter. I think it'd just be a bit more endorsement to recommend people use it even if they're just implementing core + item-search.

I think we want to at least recommend, if not require it -- it's just good practice, and we already have an openapi schema they can use for it.

I don't understand "if you wanted to include those right now you're supposed to fully implement the features API. " -- where is that required?

m-mohr:
Those two rel types originate in OGC APIs and you have to implement them for being OGC API compliant.

For STAC core, which may just be a browsable catalog and not even have a OpenAPI, I'd highly recommend to not require it. I could see us recommending it, but feel more just giving the option and the implementor has to choose. I don't see them used often in clients so it may only be a "human" thing to look at these. From a RESTful point of view (and we claim being RESTful), OpenAPIs should actually not even be a thing as everything can be just retrieved from the links.

philvarner:
> Those two rel types originate in OGC APIs and you have to implement them for being OGC API compliant.

This made me realize that we don't want this example Landing Page in core -- we should have one that is what you would see if you *only* implement core, e.g., no rel=search. We should also have these for the other 2, and then a 4th example of what a complete API that implements all 3 would look like.

Issue #146, opened 2021-06-01
Add better documentation around syntax and semantics of datetime


philvarner:
1. datetime with an instant (e.g., `1985-04-12T23:20:50.52Z`) means exact match
2. datetime interval syntax should support `..` or empty string, e.g. `/1985-04-12T23:20:50.52Z`
3. datetime RFC 3339 description and examples, because this is not clearly documented elsewhere and implementations keep getting it wrong. 

See:
- https://github.com/stac-utils/stac-pydantic/issues/78
- https://github.com/stac-utils/stac-pydantic/issues/79
- https://github.com/stac-utils/stac-pydantic/issues/80

PR #162, opened 2021-07-01
add detailed description of datetime parameter use

philvarner:
**Related Issue(s):** #146

**Proposed Changes:**

1. provide additional details about implementing the datetime parameter beyond what is provided in OAFeat.

**PR Checklist:**

### DIFF
--- a/implementation.md
+++ b/implementation.md
@@ -30,8 +30,8 @@ permitting the `If-Match` request header.
 
 The datetime parameter used by the Item Search and Features conformance classes allows the same values as the 
 [OAF datetime](http://docs.opengeospatial.org/is/17-069r3/17-069r3.html#_parameter_datetime) parameter. This allows for 
-either a single [RFC 3339](https://datatracker.ietf.org/doc/html/rfc3339) datetime or an open or closed interval that also 
-uses RFC 3339 datetimes.
+either a single [RFC 3339](https://datatracker.ietf.org/doc/html/rfc3339) datetime (e.g., `1990-12-31T23:59:59Z`) or a closed or single open ended interval that also 
+uses RFC 3339 datetimes (e.g., `1990-12-31T23:59:59Z/..`).
 
 RFC 3339 is a profile of ISO 8601, adding these constraints to the allowed values:
 - a complete representation of date and time (fractional seconds optional).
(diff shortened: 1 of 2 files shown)


implementation.md
```diff
@@ -42,7 +42,7 @@ RFC 3339 is a profile of ISO 8601, adding these constraints to the allowed value
 A simple regex for an RFC 3339 datetime is:
 
 ```
-^(\d\d\d\d)\-(\d\d)\-(\d\d)(T|t)(\d\d):(\d\d):(\d\d)(\.\d+)?(Z|([-+])(\d\d):(\d\d))$
+^(\d\d\d\d)\-(\d\d)\-(\d\d)(T|t)(\d\d):(\d\d):(\d\d)([,.]\d+)?(Z|([-+])(\d\d):(\d\d))$
```
m-mohr, line 45: AFAIK the dot needs to be escaped, otherwise it matches all characters.

philvarner: Inside `[]`, `.` is a literal and not a metacharacter like it is normally.

m-mohr: Nice, still learning new stuff about regexes.

philvarner: I did this notebook a couple of months ago to help out my understanding -- I can never remember to rules. https://github.com/philvarner/old-possum/blob/main/scala/RegexExamples.ipynb It's scala, but the strings are standard regex. 

One thing I found out because of a bug is that `-` in `[]` is a range character if it's between two other characters, but a literal if it's the first character or if it's escaped. So, I always put it as the first character if I'm matching on it, e.g., `[.-_]` matches the ascii range between `.` and `_` but `[-._]` matches those three literal chars. That's why I wrote -+ in the timezone offset above instead of +-

m-mohr: Yeah, I knew about the dash issue. Thanks for the link.

implementation.md
```diff
@@ -25,3 +25,60 @@ Implementations that support the Transactions Extension or require credentials f
...
+
+This is not a precise regex, as it matches some strings that violate semantics. There are additional restrictions, for example, 
+the month (01-12), the day (01-31), the hour (0-24), minute (0-60), seconds (0-9), and timezone offsets.  However, the best 
+strategy is to use this regex to ensure the datetime conforms to the the RFC3339 profile and then us an ISO8601 parser to produce
+a valid datetime object from the datetime string.
+
+Thereby, the recommended process for parsing the datetime is:
+
+1. uppercase the string (this avoids needing to match on both (t|T) and (z|Z))
+2. split the string on `/` to determine if it is a single datetime or an interval
```
m-mohr, line 56: The interval semantic was not mentioned before and it's not 100% clear whether the regexp accounts for this case (it does not, but should it?)

implementation.md
```diff
@@ -25,3 +25,60 @@ Implementations that support the Transactions Extension or require credentials f
...
+3. For each value of the split, check if it is either an open interval (the empty string or `..`), or if it matches the RFC3339 datetime regex
+5. ISO8601 parse datetime strings using a library such as [pyiso8601](https://github.com/micktwomey/pyiso8601) or
```
m-mohr, line 58: Maybe list two libs for different languages instead of two libs for the same language?

implementation.md
```diff
@@ -25,3 +25,60 @@ Implementations that support the Transactions Extension or require credentials f
...
+
+- 1990-12-31T23:59:59Z (no fractional seconds, Z timezone)
+- 1990-12-31T23:59:23.123Z (fractional seconds, Z timezone)
+- 1996-12-19T16:39:57-08:00 (no fractional seconds, negative offset timezone)
+- 1937-01-01T12:00:27.87+01:00 (fractional seconds, positive offset timezone)
+- 1985-04-12t23:20:50.5202020z (lowercase t for separator, lowercase z for timezone)
+
+These are several examples of datetime intervals:
+
+- `/`
```
m-mohr, line 75: What's the meaning of this?

philvarner: now that I read this again (http://docs.opengeospatial.org/is/17-069r3/17-069r3.html#_parameter_datetime), open start and open end is invalid -- I'll correct that.  

As for the empty strings, the ABNF says:

interval-open-start = [".."] "/" date-time
interval-open-end   = date-time "/" [".."]

and then later:

"Open ranges in time intervals at the start or end are supported using a double-dot (..) **or an empty string** for the start/end."

implementation.md
philvarner: I'm clarified the text. The regex is only for parsing a single datetime value.

implementation.md
philvarner: I've added one that looks good for JavaScript. I mostly didn't want to recommend something that I wasn't sure if it worked or not. The main point was the Python's built-in library doesn't work, but I don't know about any other languages. I'm not even sure the Scala/JVM one worked correctly.

cholmes (approved):
Looks great. A nice addition.

matthewhanson (approved):

Issue #158, opened 2021-06-16
Filter STAC conformance class for our sub-set of CQL?


cholmes:
We had originally envisioned that we'd be able to just 'use' CQL, but it ends up not having the right granularity for us. We've given our feedback and it likely will evolve to fit well for us.

But I've been thinking we may not have to just wait. We could make our own STAC conformance class that specifies the exact subset of CQL that we think makes sense. Initially this would just describe the parts of CQL that are necessary to support, referring to the core CQL spec. But as CQL evolves it'd simplify ours, as we could just refer to the 3 CQL classes.

With Features API we'd hopefully be able to entirely drop our 'stac cql features api' conformance class. With STAC Item Search we always need a class to 'bind' the CQL language to that endpoint. So I think from an implementors standpoint it'd mostly be the same - same set of things you implement for core CQL, though the exact conformance classes to show that may evolve a bit. This would be more work for us, to specify it more fully, but I think would let us have a more solid CQL-based Filter that works well to use earlier.

PR #163, opened 2021-07-05
change: refactor Filter Extension conformance classes

philvarner:
**Related Issue(s):** #158

**Proposed Changes:**

Refactors Filter Extension conformance classes to more closely align with OGC API Part 3 changes coming soon. 

The current draft Simple CQL class is definitely going away (https://github.com/opengeospatial/ogcapi-features/issues/579#issuecomment-874131446), but I don't think the proposed new ones meet the needs of STAC API either 🤷🏻 I think we're going to need to keep our own that don't align with the OAF ones, but hopefully they'll align eventually.  Since Part 3 is a work in progress, having our own conformance classes will help insulate us from those changes and allow us to move forward without waiting for those to be finalized.

**PR Checklist:**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 - Query extension not deprecated; recommendation to use Filter (https://github.com/radiantearth/stac-api-spec/pull/157)
+- Filter Extension conformance classes refactored to better align with STAC API use cases.
 
 ### Deprecated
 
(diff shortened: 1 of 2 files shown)


m-mohr:
I feel like we should provide a mapping between conformance classes from OGC API and STAC API, as we do with the other conformance classes, too. Not providing any OGC CQL conformance classes will break compatibility between OGC tooling and STAC tooling completely. With the mapping implementors could more easily express what they support for both worlds. That may lead to a subset of comliancy only as you'll only add conformance classes from OGC that you fully support, but overall at least some compliance could be achieved.

cholmes (approved):
Overall looks good. I'd say the one thing to add somewhere is an overview of alignment / discussion w/ OGC. I saw that you decided to go a bit of a different route with keeping the basic spatial / time ones. It'd be good just give an overview of where we stand and where OGC stands and where/how we see them coming together. This could just be in a ticket or something, and linked to from this page, or on this page.

Issue #142, opened 2021-05-26
Document link relations 


philvarner:
Add a "Relation types" section to the STAC API - Features docs similar to what we have for Using Relation Types in the core best practices. I think it would be helpful for backend and client implementors to have those collected in one place.

PR #176, opened 2021-07-20
add link relation table to each conformance class

philvarner:
**Related Issue(s):** #142

**Proposed Changes:**

1. Add table with Link Relations to each conformance class to clarify which ones should be implemented.

**PR Checklist:**

### DIFF
--- a/core/README.md
+++ b/core/README.md
@@ -1,7 +1,7 @@
 # STAC API - Core Specification
 
 - [STAC API - Core Specification](#stac-api---core-specification)
-  - [Recommended Link Relations at `/`](#recommended-link-relations-at-)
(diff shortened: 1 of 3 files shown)


cholmes (approved):

matthewhanson:
Why have the duplicated list of link relations that should be at the root catalog listed under the other endpoints?

philvarner:
> Why have the duplicated list of link relations that should be at the root catalog listed under the other endpoints?

The idea was to make them all independent and self-contained definitions. I think what i need to do is take the common ones that apply to all of them and put them somewhere else, and then reference them, then, only the ones specific to each class need to go into the api definition.

Issue #134, opened 2021-05-17
Add Collections API as a separate spec and conformance class


philvarner:
Define a separate endpoint for Collections (/collections)

Issue for PR https://github.com/radiantearth/stac-api-spec/pull/71/files

PR #179, opened 2021-07-21
add collections api definition

philvarner:
**Related Issue(s):** #134

**Proposed Changes:**

1. Add Collections conformance class defintion

**PR Checklist:**

### DIFF
--- a/collections/README.md
+++ b/collections/README.md
@@ -7,7 +7,7 @@
 	- [Conformance](#conformance)
 
 - **OpenAPI specification:** Missing
-- **Conformance URI:** <http://stacspec.org/spec/api/1.0.0-beta.1/extensions/collections>
+- **Conformance URI:** <http://stacspec.org/spec/api/1.0.0-beta.2/extensions/collections>
 - **Extension [Maturity Classification](../extensions.md#extension-maturity):** Pilot
 - **Dependencies**: [STAC API - Core](../core)
 
@@ -98,7 +98,7 @@ and if the canonical location is elsewhere it is recommended to include a `rel` 
 
 ## Conformance
 
-Any implementation that provides the STAC Collection functionality described above must add `http://stacspec.org/spec/api/1.0.0-beta.1/collections`,
+Any implementation that provides the STAC Collection functionality described above must add `http://stacspec.org/spec/api/1.0.0-beta.2/collections`,
 the conformance URI, to the `conformsTo` JSON at the landing page.
 
 The core STAC landing page (`/`) must also include a link with a `rel` of `data` that links to the `/collections` endpoint.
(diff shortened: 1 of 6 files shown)


collections/README.md
```diff
@@ -0,0 +1,106 @@
+# STAC API - Collections
+
+- [STAC API - Collections](#stac-api---collections)
+	- [Link Relations](#link-relations)
+	- [Endpoints](#endpoints)
+	- [Example](#example)
+	- [Conformance](#conformance)
+
+- **OpenAPI specification:** Missing
+- **Conformance URI:** <http://stacspec.org/spec/api/1.0.0-beta.1/extensions/collections>
```
matthewhanson, line 10: should be 1.0.0-beta.2

collections/README.md
```diff
@@ -0,0 +1,106 @@
...
+
+**NOTE**: *This conformance class is directly based on the [Features Collection](http://docs.opengeospatial.org/is/17-069r3/17-069r3.html#_collections_)
+section of OAFeat, which is in the process of becoming a 'building block in [OGC API - Common - Part 2: Geospatial 
+Data](http://docs.opengeospatial.org/DRAFTS/20-024.html) as the [Collections requirements 
+class](http://docs.opengeospatial.org/DRAFTS/20-024.html#rc_collections-section). Once the Common version is released we will 
+aim to align with it. But it still seems to be in flux.*
+
+## Link Relations
+
+The following Link relations should exist in the Landing Page (root).
```
matthewhanson, line 26: Was this intended to have Link relations for the Landing page here on the Collections endpoint?

collections/README.md
```diff
@@ -0,0 +1,106 @@
...
+			"links": [
+				{
+					"rel": "self",
+					"type": "application/json",
+					"href": "https://myservice.com/collections/cool-data"
+				},
+			],
+		}
+	],
+	"links": [
```
matthewhanson, line 81: So no other links required: root, parent?

matthewhanson:
This seems fine, but wondering what is it that isn't captured in the existing `http://www.opengis.net/spec/ogcapi-features-1/1.0/conf/core` conformance class for the /collections endpoint that requires a STAC specific Collections conformance class?

collections/README.md
philvarner: yes. We probably also need to clarify which link relations should exist on the /collections and /collections/{collectionid} endpoints for both this and ogcapi-features.

collections/README.md
philvarner: I'll improve this example.

m-mohr:
> This seems fine, but wondering what is it that isn't captured in the existing `http://www.opengis.net/spec/ogcapi-features-1/1.0/conf/core` conformance class for the /collections endpoint that requires a STAC specific Collections conformance class?

It indicated that the collections are STAC compliant...

matthewhanson (approved):

m-mohr:
Interesting PR, would like to review it as it's what we pioneered in the openEO API, but will likely take me some days to go through it.

philvarner:
@matthewhanson This was Chris's idea, and it's basically to allow a user to advertise STAC-compliant Collections without needing to implement all of OGC API Features, e.g., they only want to implement Collections and Item Search. Ideally, there would be a way to define this without duplicating it from OAFeat, but this is what we have for now.

PR #181, opened 2021-07-23
add extensions section to ocgapi-features, cleanup other extensions s…

philvarner:
**Related Issue(s):** #180

**Proposed Changes:**

1. Add Extensions section to STAC API - Features definition. I'm not sure this was linked to from anywhere prior.

**PR Checklist:**


cholmes (dismissed):
This looks good. Note that we also probably at some point should define our OGC API - Features extensions for sort and fields. Since I'd guess many STAC Servers actually use those. At some point in the future we'll switch over to what OGC is doing, but it'd be good to officially bind our 'fragments' to the features endpoints, as that was the whole idea of defining them as 'fragments', we just never got to actually making the real ogcapi-features extensions versions of them.

matthewhanson (approved):

Issue #165, opened 2021-07-09
OAFeat rendered OpenAPI document v1.0.0-beta.2 displays an error


alainbodiguel:
When accessing https://api.stacspec.org/v1.0.0-beta.2/ogcapi-features/ an error is displayed instead of the actual API description.

philvarner:
@m-mohr you mentioned this right before the mtg ended

philvarner:
@alainbodiguel thanks for filing this issue. After a few hours of working on it, I finally fixed the build and I think we'll be able to publish a new version of this.

PR #185, opened 2021-08-02
fix redoc build and cleanup warnings

philvarner:
**Related Issue(s):** #165

**Proposed Changes:**

1. change redoc build to use openapi-cli instead of swagger-cli, which should fix the issue with the published version of 
2. replace index.html with current recommended content from ReDoc
2. cleanup redoc warnings

**PR Checklist:**

### DIFF
--- a/build/index.html
+++ b/build/index.html
@@ -20,7 +20,12 @@
 </head>
 
 <body>
-  <redoc spec-url='openapi.yaml'></redoc>
+  <redoc 
+    spec-url='openapi.yaml'
+    expandResponses='200,201,202,203,204'
+    pathInMiddlePanel='true'
+    hideDownloadButton='true'
+  ></redoc>
   <script src="https://cdn.jsdelivr.net/npm/redoc@next/bundles/redoc.standalone.js"> </script>
 </body>
 
--- a/build/index.html
+++ b/build/index.html
@@ -1,20 +1,27 @@
-<html lang="en">
- <head>
-   <meta charset="utf-8">
-   <meta name="viewport" content="width=device-width,initial-scale=1">
-   <meta http-equiv="x-ua-compatible" content="ie=edge">
-   <link rel="shortcut icon" href="../assets/images/favicon.png">
-   <title>STAC API</title>
- </head>
- <body>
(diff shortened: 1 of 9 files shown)


package.json
```diff
@@ -5,7 +5,7 @@
   "repository": "https://github.com/radiantearth/stac-api-spec",
   "license": "Apache-2.0",
   "scripts": {
-    "serve": "redoc-cli serve core/openapi.yaml --watch --options.expandResponses \"200,201,202,203,204\" --options.pathInMiddlePanel true",
+    "serve": "redoc-cli serve $npm_config_class/openapi.yaml --watch --options.expandResponses \"200,201,202,203,204\" --options.pathInMiddlePanel true",
```
philvarner, line 8: syntax is now `npm run server --class=core` (where class can be core, ogcapi-features, item-search)

matthewhanson (approved):

build/index.html
```diff
@@ -1,20 +1,27 @@
...
-   <meta http-equiv="x-ua-compatible" content="ie=edge">
-   <link rel="shortcut icon" href="../assets/images/favicon.png">
-   <title>STAC API</title>
- </head>
- <body>
-   <div id="content"></div>
-   <script src="https://cdn.jsdelivr.net/npm/redoc@next/bundles/redoc.standalone.js"> </script>
-   <script type="text/javascript">
-     Redoc.init('openapi.yaml', {
-       expandResponses:"200,201,202,203,204",
```
m-mohr, line 14: Is it intentional that you drop those settings?

package.json
m-mohr: Does that work on all major OS (looking at Windows)?

philvarner: It's part of NPM rather than shell, so it should.

build/index.html
philvarner: I just readded them. I'm not sure why we made these changes to the default, but I also didn't test them to see how they were different. I'll try that out before changing.

m-mohr: I made them, they give a bit better UI experience IMHO.

PR #188, opened 2021-08-03
fix for v1.0.0-beta.2 ogcapi-features

philvarner:
**Related Issue(s):** #165

**Proposed Changes:**

1. manual fix for the page https://api.stacspec.org/v1.0.0-beta.2/ogcapi-features/
This is automatically built from circleci when a tag is created. I can't figure out how to get this to publish locally from master, so this is just adding the files manually.
2. The CircleCI build error is b/c this branch doesn't have a circleci build config, because it doesn't need one.
3. I spot-checked several of the changes in the openapi configs, and they equivalent code -- in most cases, they're a change from definitions being inline to being referenced.

**PR Checklist:**

### DIFF
--- a/v1.0.0-beta.2/ogcapi-features/extensions/transaction/index.html
+++ b/v1.0.0-beta.2/ogcapi-features/extensions/transaction/index.html
@@ -1,20 +1,32 @@
-<html lang="en">
- <head>
-   <meta charset="utf-8">
-   <meta name="viewport" content="width=device-width,initial-scale=1">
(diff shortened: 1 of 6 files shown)


philvarner:
1. The CircleCI build error is b/c this branch doesn't have a circleci build config, because it doesn't need one.
2. I spot-checked several of the changes in the openapi configs, and they equivalent code -- in most cases, they're a change from definitions being inline to being referenced.

Issue #195, opened 2021-08-11
STAC API - Collections feedback


m-mohr:
Some feedback for the STAC API - Collections: 
1. The Maturity says "Pilot", which is a bit weird as there's no implementation for that conformance class yet (I think) and no OpenAPI document is available. I guess it should be "Proposal" instead?
2. As this is explicitly for STAC API - Collections, I think the data relation type should be required. There are no child links in OAFeat defined as far as I know, so I think the child relation should just not be listed or at least not be listed in the main table of relation types to implement. (Just seeing that it's required at the end in the Conformance section, but listed as "should" in the table. That conflicts a bit...)
3. Is a root relation type really useful in the root "catalog" / landing page? It seems to mirror self 1:1.
4. The items relation type should probably point to the full OGC API Features extension so that people see what "items" actually should be. 
5. Isn't the list of endpoints missing the `/` endpoint (and potentially an indication on the API (e.g. `/api`) endpoint)?
6. I'd propose to put the example into a separate file like we do it in most other cases, too. Then it could also be extended a bit with an additional collection in the current example and an additional example for the landing page.
7. The relation types mentioned in the example text (canonical, alternate representation) are not listed in any of the tables.
8. The "NOTE: When this aligns with OGC API - Common it will also need to add it to the /conformance endpoint." should probably be phrased differently. It basically also applies to Features and for better OGC API compliance I'd actually add it to the top list of relation types. I don't see this is not (directly) inherited from OGC APIs while the other parts are, e.g. the API doc endpoint(s).

philvarner:
>     1. The Maturity says "Pilot", which is a bit weird as there's no implementation for that conformance class yet (I think) and no OpenAPI document is available. I guess it should be "Proposal" instead?

I put it in pilot because it's just a subset of `STAC API - Features`. Any impl that already supports that class can just add this conformance class.

> 2. As this is explicitly for STAC API - Collections, I think the data relation type should be required. There are no child links in OAFeat defined as far as I know, so I think the child relation should just not be listed or at least not be listed in the main table of relation types to implement. (Just seeing that it's required at the end in the Conformance section, but listed as "should" in the table. That conflicts a bit...)

The data relation is required. We follow the RFC meaning of "should", which means "must".

> 3. Is a root relation type really useful in the root "catalog" / landing page? It seems to mirror self 1:1.

I kept this just to be consistent with all the other APIs. 

[...]

PR #200, opened 2021-08-18
Shall and endpoint table

philvarner:
…implemented for each conformance class

**Related Issue(s):** #195 

**Proposed Changes:**

1. change wording of required endpoints for each conformance class from "should" to "shall".

Technically this is a breaking change, but it was understood before that these were required, we just used the wrong word in the spec.

**PR Checklist:**

- [ ] ??? This PR has **no** breaking changes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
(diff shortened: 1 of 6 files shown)


cholmes (approved):
Looks good. Do we define somewhere our use of should / shall? Like refer to a document or explain it somewhere?

philvarner:
We don't have it explicitly stated anywhere, but I think it's widely known enough that we don't need to.

Issue #196, opened 2021-08-17
Minor rework of Filter Ext conformance classes, and mapping table


philvarner:
Rework Filter Ext conformance classes to match current latest Part 3, primarily that temporal comparisons are part of Basic.

Issue #194, opened 2021-08-11
OGC API conformance classes for Filter


m-mohr:
To improve compatibility, I'd propose to also list the OGC API conformance classes for the filter extensions whenever one of our conformance classes fully includes one of theirs or if a set of our conformance classes map one of their conformance classes. This would help to make out ecosystem align more. I don't see a good reason (yet?) why we stick solely with custom conformance classes instead of mapping them whenever possible. In general, it would be great to have a list of how our conformance classes relate to the OGC API conformance classes.

philvarner:
OGC API conformance classes draft updated https://github.com/opengeospatial/ogcapi-features/commit/29ab01f447d44b3d843704a31258b02be82ecdfe

philvarner:
I ended up not adding a table, since we're not aligned except for one minor difference

m-mohr:
I don't understand. My point was to add OGC API conformance URIs and I don't see them in the PR. Also, why are we completely misaligned with them?

philvarner:
I typed that too quickly -- I meant to say were **are** aligned except for one minor difference (upper/lower function in Advanced Comparison Operators). All of the conformance classes are named the same, and all of the URIs have the same ending.

m-mohr:
Ah, okay. But wouldn't it be more clear to just list them so that people can copy&paste them easily instead of have them search for them? From the docs it doesn't become directly clear how to "convert" the URLs (to get OGC API compliant), too, right?

Issue #139, opened 2021-05-20
Add more advanced query examples for Filter Extension


philvarner:
Add more advanced query examples for Filter Extension. Right now, there are just a few logical operations, but they can be arbitrarily complex.

- multiple non-intersecting polygons
- landsat:data_coverage vs. sentinel:data_coverage (with IS NULL)

PR #202, opened 2021-08-23
Filter Extension - Conformance classes realign

philvarner:
**Related Issue(s):** 
* #196 
* #194
* #139

**Proposed Changes:**

1. Align Filter Extension conformance classes with OAFeat (details in changelog)

One difference with OAFeat CQL2 is that we do not require upper/lower as part of Advanced Comparison Ops, though I am reconsidering that. Now that IS NULL is part of the Basic CQL, LIKE is the only operator that can't otherwise be expressed in Basic CQL (e.g., IN and BETWEEN can be converted to more verbose clauses).

I would prefer that an implementor didn't need to implement function parsing to support this class, but I think, ultimately, it's not worth having this minor difference between STAC API and OAFeat CQL2, so I think we should defer to what OAFeat finalizes.

**PR Checklist:**

### DIFF
--- a/extensions.md
+++ b/extensions.md
@@ -79,7 +79,7 @@ the service supports. This are listed at the top of each extension description, 
   - <https://api.stacspec.org/v1.0.0-beta.4/item-search#filter:temporal-operators>
   - <https://api.stacspec.org/v1.0.0-beta.4/item-search#filter:functions>
   - <https://api.stacspec.org/v1.0.0-beta.4/item-search#filter:arithmetic>
-  - <https://api.stacspec.org/v1.0.0-beta.4/item-search#filter:array-operators-operators>
+  - <https://api.stacspec.org/v1.0.0-beta.4/item-search#filter:array-operators>
   - <https://api.stacspec.org/v1.0.0-beta.4/item-search#filter:property-property>
 - [Context](item-search/README.md#context)
   - <https://api.stacspec.org/v1.0.0-beta.4/item-search#context>
--- a/extensions.md
+++ b/extensions.md
@@ -106,6 +106,7 @@ Please contact a STAC maintainer or open a Pull Request to add your extension to
 | Name                                                                | Scope                               | Description                                                     | Vendor                                         |
 | ------------------------------------------------------------------- | ----------------------------------- | --------------------------------------------------------------- | ---------------------------------------------- |
 | [Free-text Search](https://github.com/cedadev/stac-freetext-search) | [Item Search](item-search/) request | Adds `q` parameter and free-text search against item properties | [CEDA, STFC, UKRI](https://github.com/cedadev) |
+| [Context Collections](https://github.com/cedadev/stac-context-collections) | [Item Search](item-search/) request | Adds a `collections` keyword to the [context](https://github.com/radiantearth/stac-api-spec/tree/master/fragments/context) extension response. | [CEDA, STFC, UKRI](https://github.com/cedadev) |
 
 ## Creating new extensions
 
--- a/extensions.md
+++ b/extensions.md
@@ -67,6 +67,7 @@ the service supports. This are listed at the top of each extension description, 
 
 - [Fields](item-search/README.md#fields)
   - <https://api.stacspec.org/v1.0.0-beta.4/item-search#fields>
+  - <https://api.stacspec.org/v1.0.0-beta.4/ogcapi-features#fields>
 - [Filter](item-search/README.md#filter)
   - <http://www.opengis.net/spec/ogcapi-features-3/1.0/conf/filter>
   - <http://www.opengis.net/spec/ogcapi-features-3/1.0/conf/basic-cql>
@@ -83,8 +84,10 @@ the service supports. This are listed at the top of each extension description, 
   - <http://www.opengis.net/spec/ogcapi-features-3/1.0/conf/property-property>
 - [Context](item-search/README.md#context)
   - <https://api.stacspec.org/v1.0.0-beta.4/item-search#context>
+  - <https://api.stacspec.org/v1.0.0-beta.4/ogcapi-features#context>
 - [Sort](item-search/README.md#sort)
   - <https://api.stacspec.org/v1.0.0-beta.4/item-search#sort>
+  - <https://api.stacspec.org/v1.0.0-beta.4/ogcapi-features#sort>
 - [Transaction](ogcapi-features/extensions/transaction/README.md)
   - <https://api.stacspec.org/v1.0.0-beta.4/ogcapi-features/extensions/transaction>
 - [Items and Collections API Version](ogcapi-features/extensions/version/README.md)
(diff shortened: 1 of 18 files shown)


extensions.md
```diff
@@ -69,10 +69,18 @@ the service supports. This are listed at the top of each extension description,
   - <https://api.stacspec.org/v1.0.0-beta.4/item-search#fields>
 - [Filter](item-search/README.md#filter)
   - <https://api.stacspec.org/v1.0.0-beta.4/item-search#filter:filter>
-  - <https://api.stacspec.org/v1.0.0-beta.4/item-search#filter:simple-cql>
+  - <https://api.stacspec.org/v1.0.0-beta.4/item-search#filter:basic-cql>
```
philvarner, line 72: all of these names align with the OGAFeat CQL2 names.

fragments/filter/README.md
```diff
@@ -768,13 +756,13 @@ This uses the same queryables as Example 4.
...
-The only temporal operator required is `ANYINTERACTS`, which follows the same semantics as the existing 
-`datetime` parameter. This is effectively that the datetime or interval operands have any overlap between them.
+The only temporal operator required is `ANYINTERACTS`. This is effectively that the datetime or interval operands 
+have any overlap between them.
 
 #### Example 6: ANYINTERACTS cql-text (GET)
 
 ```javascript
-/search?filter=datetime ANYINTERACTS 2020-11-11
+/search?filter=datetime ANYINTERACTS 2020-11-11T00:00:00Z/2020-11-12T00:00:00Z
```
philvarner, line 765: these examples were just wrong

fragments/filter/README.md
```diff
@@ -230,16 +224,15 @@ not compliant with this extension.
...
 returned according to OAFeat Part 3. It is recognized that this is a severe restriction in STAC APIs that have highly variable 
-and dynamic content. It is possible that this will change in the OAFeat Part 3 spec, see 
-this [issue](https://github.com/opengeospatial/ogcapi-features/issues/582). For now, implementers may choose to allow 
-fully-qualified property 
-names not advertised in queryables (e.g., `properties.eo:cloud_cover`, in anticipation that there be some
-mechanism to explicitly allow or advertise this in the future.
+and dynamic content, so this behavior may be modified by setting the `additionalProperties` attribute in the
+queryables definition to `true`.  As such, any syntactically-valid term for a property will be accepted, and the
+matching semantics are such that, if an Item does not have an attribute by that name, the value is assumed to be
+`null`.  It is recommended to use fully-qualified property names (e.g., `properties.eo:cloud_cover`).
```
m-mohr, line 235: Should we include something to explain that you can implement "simplified" access to complicated nested structures? So that you could add a queryable "common_band_name" that searches for a common name in eo:bands in assets?

fragments/filter/README.md
```diff
@@ -507,11 +491,12 @@ OGC API Features filters only operate against a single collection already.
...
 
 ```javascript
-GET /search?filter=collection = 'landsat8_l1tp' 
+GET /search?filter=collection = 'landsat8_l1tp'
   AND gsd <= 30
-  AND eo:cloud_cover <= 10 
-  AND datetime ANYINTERACTS 2021-04-08T04:39:23Z/2021-05-07T12:27:57Z 
-  AND INTERSECTS(geometry, POLYGON((43.5845 -79.5442, 43.6079 -79.4893, 43.5677 -79.4632, 43.6129 -79.3925, 43.6223 -79.3238, 43.6576 -79.3163, 43.7945 -79.1178, 43.8144 -79.1542, 43.8555 -79.1714, 43.7509 -79.6390, 43.5845 -79.5442)) 
+  AND eo:cloud_cover <= 10
+  AND datetime >= "2021-04-08T04:39:23Z"
```
m-mohr, line 497: Wouldn't this usually be covered by a BETWEEN?

fragments/filter/README.md
m-mohr: What does this mean? They still need to be changed here?

fragments/filter/README.md
```diff
@@ -762,79 +718,153 @@ This uses the same queryables as Example 4.
...
+      }
+    ]
+  }
+}
+```
+
+### Example 8: Spatial Disjunction
+
+One limitation of the `intersects` parameter is that only a single geometry may be provided. While most
+GeoJSON geometries can be combined to form a composite (e.g., multiple Polygons can be combined to form a
```
m-mohr, line 795: There's also the (discouraged?) GeometryCollection...

fragments/filter/README.md
```diff
@@ -762,79 +718,153 @@ This uses the same queryables as Example 4.
...
+One of the main use cases for STAC API is doing cross-collection query. Commonly, this means that items have
+different sets of properties. For example, a collection of Sentinel 2 data may have a property
+`sentinel:data_coverage` and a collection of Landsat 8 data may have a corresponding property
+`landsat:coverage_percent`, both representing what percentage of a given gridded scene actually contains
+data.  However, we many also want to also include in our result items that do not have a value defined for
+either of those properties. 
+
+#### Example 9: cql-text (GET)
+
+```javascript
```
m-mohr, line 849: Why are these snippets JavaScript? I guess ``http`` is more correct? (you may also want to include GET in front of the call?)

implementation.md
```diff
@@ -70,7 +70,8 @@ RFC 3339 datetime or an interval) is:
    language standard libraries do not parse ISO8601 datetimes correctly, for example, the built-in
    Python datetime library does not handle `Z` as a timezone.
 
-Below are a few examples of valid RFC 3339 datetimes. Note the uses of fractional seconds, the use of `.` as the fractional seconds separator, Z or z as a timezone, 
+Below are a few examples of valid RFC 3339 datetimes. Note the uses of fractional seconds, the use of `.` 
+as the fractional seconds separator, Z or z as a timezone,
```
m-mohr, line 74:
```suggestion
as the fractional seconds separator, Z (recommended) or z as a timezone, 
```

implementation.md
```diff
@@ -70,7 +70,8 @@ RFC 3339 datetime or an interval) is:
    language standard libraries do not parse ISO8601 datetimes correctly, for example, the built-in
    Python datetime library does not handle `Z` as a timezone.
 
-Below are a few examples of valid RFC 3339 datetimes. Note the uses of fractional seconds, the use of `.` as the fractional seconds separator, Z or z as a timezone, 
+Below are a few examples of valid RFC 3339 datetimes. Note the uses of fractional seconds, the use of `.` 
+as the fractional seconds separator, Z or z as a timezone, 
 positive and negative arbitrary offset timezones, and T or t as a separator between date and time. While
```
m-mohr, line 75:
```suggestion
positive and negative arbitrary offset timezones, and T (recommended) or t as a separator between date and time. While 
```

m-mohr (approved):
Some comments, but nothing really blocking I think.
I love that this is now fully aligned!

fragments/filter/README.md
philvarner: cut and paste. I'll change them to have no language.

fragments/filter/README.md
philvarner: there's a section about that later in the doc, around line 300.

fragments/filter/README.md
philvarner: It would, but BETWEEN is in the Advanced Comparison Operators conformance class, and I wanted these first examples to only use the Basic operators.

fragments/filter/README.md
philvarner: I fixed them. I was stating that the reason I fixed them was that they were just invalid expressions(e.g., `datetime ANYINTERACTS 2020-11-11` isn't valid because it's just a date and not a datetime range), rather than there was actually a change to the syntax.

PR #210, opened 2021-09-09
incorporate feedback on collections from beta.3, cleanup related text

philvarner:
incorporate feedback on collections from beta.3, cleanup related text in other components

**Related Issue(s):** #195

**Proposed Changes:**

1. incorporate feedback from @m-mohr on collections
2. fix more instances of should vs. shall/must

**PR Checklist:**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Support binding Sort, Fields, and Context Extensions to STAC Features items resource
   endpoint (`/collections/{collection_id}/items`)
 - In Collections, added `canonical` rel type, added `/` and `/api` to list of endpoints
-- In Item Search, added Endpoint table
+- In Item Search, added endpoint table
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Support binding Sort, Fields, and Context Extensions to STAC Features items resource
   endpoint (`/collections/{collection_id}/items`)
+- In Collections, added `canonical` rel type, added `/` and `/api` to list of endpoints
+- In Item Search, added Endpoint table
 
 ### Changed
 
--- a/collections/README.md
+++ b/collections/README.md
@@ -1,12 +1,10 @@
 # STAC API - Collections
 
 - [STAC API - Collections](#stac-api---collections)
-	- [Link Relations](#link-relations)
-	- [Endpoints](#endpoints)
-	- [Example](#example)
-	- [Conformance](#conformance)
+  - [Endpoints](#endpoints)
+  - [Example](#example)
 
-- **OpenAPI specification:** Missing
+- **OpenAPI specification:** [openapi.yaml](openapi.yaml) ([rendered version](https://api.stacspec.org/v1.0.0-beta.4/collections))
 - **Conformance URI:** <http://stacspec.org/spec/api/1.0.0-beta.4/extensions/collections>
 - **Extension [Maturity Classification](../extensions.md#extension-maturity):** Pilot
 - **Dependencies**: [STAC API - Core](../core)
@@ -21,8 +19,6 @@ Data](http://docs.opengeospatial.org/DRAFTS/20-024.html) as the [Collections req
 class](http://docs.opengeospatial.org/DRAFTS/20-024.html#rc_collections-section). Once the Common version is released we will 
 aim to align with it. But it still seems to be in flux.*
 
-## Link Relations
-
 The following Link relations shall exist in the Landing Page (root).
 
 | **rel**        | **href**             | **From**       | **Description**                                                                                                                                                         |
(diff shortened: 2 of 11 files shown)


CHANGELOG.md
```diff
@@ -11,6 +11,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Support binding Sort, Fields, and Context Extensions to STAC Features items resource
   endpoint (`/collections/{collection_id}/items`)
+- In Collections, added `canonical` rel type, added `/` and `/api` to list of endpoints
+- In Item Search, added Endpoint table
```
m-mohr, line 15:
```suggestion
- In Item Search, added endpoint table
```

collections/README.md
```diff
@@ -32,115 +28,116 @@ The following Link relations shall exist in the Landing Page (root).
...
 ## Endpoints
 
-| Endpoint                      | Returns    | Description                                                          |
-| ----------------------------- | ---------- | -------------------------------------------------------------------- |
-| `/collections`                | JSON       | Object with a list of Collections contained in the catalog and links |
-| `/collections/{collectionId}` | Collection | Returns single Collection JSON                                       |
+| Endpoint | Returns | Description |
+| -------- | ------- | ----------- |
+| `/`                           | Catalog        | Landing Page and root Catalog |
+| `/api` (recommended)          | OAFeat OpenAPI | The OpenAPI service description |
```
m-mohr, line 74: This is a bit ambiguous. Is the whole row recommended or just the name of the endpoint? (I guess the latter?)

item-search/README.md
```diff
@@ -74,14 +75,22 @@ This link should look like:
...
 
-Implementations that support `POST` should add a second link with the same structure but with a `method` of `POST`. 
+Implementations that support `POST` should add a second link with the same structure but with a `method` of `POST`.
 
+## Endpoints
+
+| Endpoint | Returns | Description |
+| -------- | ------- | ----------- |
+| `/`                  | Catalog         | Landing Page and root Catalog |
+| `/api` (recommended) | OAFeat OpenAPI  | The OpenAPI service description |
```
m-mohr, line 85: See comment above about a similar line.

ogcapi-features/README.md
```diff
@@ -62,12 +62,12 @@ The core OGC API - Features endpoints are shown below, with details provided in
 | Endpoint                                        | Returns                                                 | Description                                                                                                                                |
 | ----------------------------------------------- | ------------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------ |
 | `/`                                             | [Catalog](../stac-spec/catalog-spec/README.md)          | Landing page, links to API capabilities                                                                                                    |
+| `/api`                                          | OpenAPI 3.0 JSON                                        | Returns an OpenAPI description of the service from the `service-desc` link `rel` - not required to be `/api`, but the document is required |
```
m-mohr, line 65: Here it doesn't say recommended?

collections/README.md
```diff
@@ -32,115 +28,116 @@ The following Link relations shall exist in the Landing Page (root).
...
 | `root`   | `/`                                 | STAC Core | The root URI                               |
 | `parent` | `/`                                 | OAFeat    | Parent reference, usually the root Catalog |
 | `self`   | `/collections/{collectionId}`       | OAFeat    | Self reference                             |
-| `items`  | `/collections/{collectionId}/items` | OAFeat    | Items within Collection                    |
+
+Additionally, these relations may exist for the `/collections/{collectionId}` endpoint:
+
+| **rel**  | **href**                            | **From**  | **Description**                            |
+| -------- | ----------------------------------- | --------- | ------------------------------------------ |
+| `canonical`        | various              | STAC Core      | The child STAC Catalogs & Collections. Provides curated paths to get to STAC Collection and Item objects |
```
m-mohr, line 63: I don't really understand what this means?!

collections/README.md
philvarner: just the name of the endpoint. I've clarified the text now by adding it to the description.

collections/README.md
philvarner: I tried to clarify. This was one of @cholmes's ideas -- I think the main use cases is federated catalogs, where the canonical url (to another catalog in another place) is different from the URL you used to retrieve the catalog.

philvarner:
@m-mohr I think I addressed all of your concerns here (thanks!)

ogcapi-features/README.md
m-mohr: Hmm, OAFeat is working on 3.1 already so we should try to be future-proof, I think.

collections/README.md
m-mohr: I think it might be easier to understand if we align with the RFC6596 description of canonical:

>    The canonical link relation specifies the preferred IRI from
   resources with duplicative content.  Common implementations of the
   canonical link relation are to specify the preferred version of an
   IRI [...]

Even in the revised version there's a lot of "canonical" in the text, which feels like a recursive definition.

Also, having 
> The child STAC Catalogs & Collections. 

at the beginning of the description is a bit confusing as it conflicts with the "child" relation, which is also used for "The child STAC Catalogs & Collections."

Maybe we should go with something like:
> Provides the preferred paths to get to STAC Collection and Item objects, if they differ from the URL that was used to retrieve the STAC object and thus duplicate other content. This can be useful in federated catalogs that present metadata that has a different location than the source metadata.

ogcapi-features/README.md
philvarner: IIRC, we were blocked for a while on some of our tooling not supporting it yet. I'll check on that again. Filed https://github.com/radiantearth/stac-api-spec/issues/213 to track updating to 3.1.

m-mohr: My intent was not to update to 3.1 here, just to make it future-proof and allow it!

collections/README.md
philvarner: sounds good to me, updated.

ogcapi-features/README.md
philvarner: Oh, I understand now. I think we should discuss that more, since I think that makes client implementation more difficult. Definitely something we should consider for the next beta/rc. https://github.com/radiantearth/stac-api-spec/issues/214

m-mohr (approved):

Issue #193, opened 2021-08-10
change what gets published to https://api.stacspec.org/v1.0.0-beta.3/


philvarner:
currently, it's core. not sure what should be there.

PR #212, opened 2021-09-17
#193 add an index to the 4 conformance classes

philvarner:
**Related Issue(s):** #193

**Proposed Changes:**

1. change the root dir for the base URL for all the conformance class URIs to have an index, rather than copying the core index to the root. 

**PR Checklist:**


build/redoc_index.html
```diff
@@ -0,0 +1,32 @@
+<!DOCTYPE html>
```
philvarner, line 1: index.html was copied to this filename

matthewhanson (approved):

Issue #222, opened 2021-10-13
Update CQL2 JSON examples per changes to OAFeat Part 3


philvarner:
These changes are expected to be finalized Oct 25. https://github.com/opengeospatial/ogcapi-features/issues/601#issuecomment-940066783

PR #225, opened 2021-10-26
Updates to Filter Extension for CQL2

philvarner:
**Related Issue(s):** #222

**Proposed Changes:**

- Filter Extension updates to align with changes to OAFeat CQL2 spec
 - Updated all "CQL" usages to "CQL2"
  - Most conformance class URIs are now prefixed with `http://www.opengis.net/spec/cql2/` instead
    of `http://www.opengis.net/spec/ogcapi-features-3/`
  - Conformance classes `http://www.opengis.net/spec/ogcapi-features-3/1.0/conf/basic-cql`, 
    `http://www.opengis.net/spec/ogcapi-features-3/1.0/conf/cql-text`, and
    `http://www.opengis.net/spec/ogcapi-features-3/1.0/conf/cql-json` have had `cql` replaced
    with `cql2` (in addition to the prefix change) to
    become `http://www.opengis.net/spec/cql2/1.0/conf/basic-cql2`,
    `http://www.opengis.net/spec/cql2/1.0/conf/cql2-text`, and
    `http://www.opengis.net/spec/cql2/1.0/conf/cql2-json`
  - Significant changes to CQL2 JSON format, now using `op` and `args` structure
  - Spatial operator `INTERSECTS` is now `S_INTERSECTS`
  - Temporal operator `ANYINTERACTS` is now `T_INTERSECTS`
  - Updated Example 3 (now Example 5) to make it clear that property to property comparisons require the
    Property-Property Comparisons conformance class
  - Added CQL2 Case Insensitive Comparison 
    (`http://www.opengis.net/spec/cql2/1.0/conf/case-insensitive-comparison`) conformance class
    that adds functions LOWER and UPPER for case-insensitive comparison

**PR Checklist:**

- [ ] This PR has **no** breaking changes. **THIS PR HAS BREAKING CHANGES**


jisantuc:
Should review here wait on the symbolic operators change upstream?

philvarner:
> Should review here wait on the symbolic operators change upstream?

My understanding is that the current published spec allows either symbolic or alphabetic, and that's changing to only allow symbolic, so these are currently valid wrt that.

m-mohr:
Quickly looked at it and looks good to me, but I leave the full review and final approval up to an implementor.

philvarner:
@bitner Added some examples for BETWEEN, IN, LIKE, UPPER, and LOWER. If you accept the invite for this repo, you may be able to approve the PR. Otherwise, @m-mohr or @cholmes, would you approve it based on @bitner's comments?

fragments/filter/README.md
```diff
@@ -932,3 +941,103 @@ filter=sentinel:data_coverage > 50 OR landsat:coverage_percent < 10 OR (sentinel
...
+
+#### Example 12: cql2-json (POST)
+
+```json
+{
+  "filter": {
+    "op": "eq", 
+    "args": [
+      { 
+        "function" : "lower",
```
bitner, line 1021: So upper and lower are implemented as "function" not with "op"?

As these are standard, shouldn't they use "op"? My impression is that "function" was for platform defined functions.

philvarner: Correct. They're not operators since they don't return a boolean value -- they return a string value. However, I looked it up in the spec and I did define them wrong -- they're not operators or functions, they're a separate thing, which is confusing, because the text syntax is exactly the same as if they were functions. I'll update the examples.

The spec is here:
https://github.com/opengeospatial/ogcapi-features/blob/b752479d850c52f372c4164202f041ecbe906e46/cql2/standard/schema/cql2.json

philvarner: I'm going to ask the CQL2 folks about this, because it seems like it would be simpler if it was just a pre-defined function.

philvarner: Filed this issue https://github.com/opengeospatial/ogcapi-features/issues/656

philvarner: ah, so this is changing in CQL2 https://github.com/opengeospatial/ogcapi-features/pull/641

UPPER/LOWER will be replace with a *function* CASEI

philvarner: After looking at CASEI, I have some significant concerns about it. I'm just going to yank it for beta.5, and re-add after it's stabilitzed in CQL2. I wrote up the concerns here if you're interested 
https://github.com/opengeospatial/ogcapi-features/pull/641#issuecomment-985685159

bitner (approved):

Issue #159, opened 2021-06-21
Browsable APIs


matthewhanson:
There's been a few implementations of browsable APIs that allow browsing of Items within a Collection. staccato implemented this at one point, CMR-STAC allows browsing of sub-catalogs in a Collection by year/month/day.  There's never been any completed proposal for this for stac-api-spec. It's conformant within stac-spec itself, because it mimics the structure of a static catalog. 

In a similar fashion, @lossyrob and I were recently discussing a browsable API endpoint for the purpose of grouping together related collections. For example, MODIS has a variety of products, how can we group those together in a catalog within an API?

I'm not sure yet if either of these use cases actually require an addition (extension) to stac-api-spec, because it really is just an implementation of a "static" catalog (albeit in a programmatic dynamic way), where any catalog within the entire structure could contain the proper conformance classes and search links and be an independent API. This can be seen in [CMR-STAC](https://cmr.earthdata.nasa.gov/stac) where the root catalog is actually just a static catalog that has a bunch of children...each of the children is it's own STAC API for a single NASA EOSDIS provider.

Thought it might be worthwhile to discuss some implementation details to work out if anything is required on the spec side. If you have worked on an implementation like this, or have some ideas about how you might implement it, please weigh in.

For case # 1 - Browsable catalogs for Items within a collection
- CMR-STAC uses CMR's aggregation endpoint to dynamically create child catalogs by year, then month, then day: ex: https://cmr.earthdata.nasa.gov/stac/USGS_EROS/collections/ETM.v1/2003/01/02 which has a child that points to the item https://cmr.earthdata.nasa.gov/stac/USGS_EROS/collections/ETM.v1/items/G162894313-USGS_EROS
- IIRC [staccato](https://github.com/planetlabs/staccato) used a dynamic approach where you could specify via URL parameters fields to create sub-catalogs of, e.g., `/collections/<cid>/path/087/row/435`.  This doesn't provide child catalogs in the same sense, but such links could easy be added in at the collection level for some default fields to browse by. @joshfix 

For case #2 - Browsable catalogs for grouping collections

CMR-STAC does this, but only by data providers, which is a concept already in CMR.
My thought was that you could have your root catalog, which would be a normal STAC API endpoint for searching your entire archive. For "child" endpoints you would group the collections dynamically via some existing parameter or a new one added via an extension (note this implies collection search is added to the spec).
[...]

Issue #138, opened 2021-05-18
'Browse' conformance class? 


cholmes:
Building on #137 it could be useful to have a conformance class that lets clients know that they can follow the 'child' links down to find items. Ie that they can fully treat this API as if it were a complete STAC Catalog (acts just like a static catalog). Ideally it would communicate to clients that they can navigate from the child links to every single Item that is contained in the API's. 

This section could discuss how to potential use aggregations to make the child browsing be a sort of faceted search. And also recommend how to use rel=canonical URL's to point back to a single canonical location.

Issue #137, opened 2021-05-18
Clarify what should go in 'child' links and if they're required, etc.


cholmes:
Most implementors get pretty confused as to what's supposed to go in 'child' links, and how that relates to the data resource to collections/. 

The child links are quite useful, as they'll be crawled and parsed by any STAC tool, but we should make a recommendation of what should go there. And decide if we want to require them, or have it be ok if they aren't used. A common approach seems to be to just have them be a reflection of the individual resources under 'collection'. Which does seem to have some value to be able to have it 'work' with any STAC tool, but we should make it clear why we duplicate things.

We also should explain how it can be used to provide different 'views' of the data, and can be used to browse all the way down to items.

Issue #17, opened 2019-04-24
Best Practices for Browsable APIs


matthewhanson:
As I was cleaning up the API Spec, I came across this:

----

An additional best practice is to use the WFS items available in `/collections/{collectionId}/items` as the "canonical" web location. Then the STAC Catalogs returned from `/stac/` can either link directly to those (from the appropriate sub-catalog - for example `/stac/landsat8/42/31/2017/` would be a catalog consisting of links to `/collections/`). Or it can return JSON in the link structure (like `/stac/landsat8/42/31/2017/item203123.json`), and have that returned JSON use a link with `rel=canonical` that goes back to the `Item` that is in the collection.

----

I removed it, but not because I think it without merit, but it's just not enough information for an implementer, and it's really not part of the spec. We've talked about this before, and the idea is basically that partial URL components that don't direct to an actual Item will return a STAC catalog result, mirroring what we see with static catalogs. For instance, http://landsat-stac.s3.amazonaws.com/landsat-8-l1/151/018/catalog.json isn't an Item, or the root catalog, or a collection, it is a sub-catalog, and if we were to visit https://sat-api.developmentseed.org/collections/landsat-8-l1/151/018 it would be nice to have it also return a similar sub-catalog so that one can "browse" the catalog.

@joshfix implemented something like this, but dynamically. Whereas the above represents a singular pre-defined view into the data,with dynamically generated "views" one could browse the API using any STAC metadata field. With the above one browses by path, then by row, then by date, but this would allow you to browse by e.g. Cloud cover and then sensor.

Currently, the STAC API however does not define what a sub-catalog is, nor does the spec define an `Item` being a child of a Catalog (or Collection). Instead Collections have a link to rel=`Items` (**not rel=`Item`!!**), which is a search that returns all items within that collection.

To do a browsable API correctly though, I think an API would also need to implement aggregations somehow. If you are browsing through metadata you will need to generate links dynamically of all the possible values for the specified metadata field. *I think* this means needing to define aggregation endpoints in order to support it, and they would be useful on their own anyway.

So I removed this paragraph because a browsable API warrants it's own discussion and more than a few sentences in the spec.

cholmes:
Moving this to beta.2. @matthewhanson and @m-mohr and I discussed this on a phone call 11-16-20, and agreed that with beta.1 we'll likely have the core spec in much better place for 'browse' API's, since we're aiming to make it very much discoverable / browsable, and the constructs with STAC core should be compatible. 

[...]

PR #229, opened 2021-11-08
Browseable recommendations and Children conformance class

philvarner:
**Related Issue(s):** #17 #159 #138 #137 #230

**Proposed Changes:**

1. Better description of what the Core conformance class means
2. Detailed descriptions of how to use sub-catalogs for browse
3. STAC API - Children conformance class to get all child Catalog and Collection metadata in one call

**PR Checklist:**


core/README.md
```diff
@@ -62,24 +72,66 @@ A `service-doc` endpoint is recommended, but not required.
 | ------------- | ----------- | -------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
 | `service-doc` | `/api.html` | OAFeat OpenAPI | An HTML service description.  Uses the `text/html` media type to refer to a human-consumable description of the service. The path for this endpoint is only recommended to be `/api.html`, but may be another path. |
 
-Additionally, `child` relations may exist to individual catalogs and collections.
+Additionally, `child` relations may exist to child Catalogs and Collections and `item` relations to Items. These
+relations form a directed acyclic graph that supports browseable traversal.
```
jisantuc, line 76: I'm throwing this comment here because it's the first mention of the acyclicality of the graph -- I don't think this is technically correct. I think the existence of `parent` links means that the graph has cycles. Moreover, I think it's _good_ that the graph has cycles -- it means we can get back up a level without using the back button / that we can retrace our steps using only the data. Is acyclicality an important property? I've thought about this a bit and I'm not sure what it gets us. I think this is just a normal directed graph.

philvarner: Ah, so that's a good point and one I should be more clear on. I was primarily thinking about the graph of `child` links, which must be a DAG. Also to make it clear that these are DAGs and not only trees, since the subcatalogs don't have to be distinct (e.g., you can slice them up by date or grid id, not only one)

jisantuc: Is that even necessarily the case? I doubt anyone has mutual `child` relationships, but:

- it's not explicitly disallowed
- I'm not aware of any notes recommending against multi-part cycles like catalog a has item a as a child, item a has catalog b as a child (nothing says items can't have children as far as I know), catalog a is a child of catalog b

I have no idea why anyone would do such a thing, I just read a nice Alloy 6 tutorial earlier today and it might have poisoned my brain forever.

Either way, unless the acyclicality is important for some reason, I'm not sure it's doing much work except encouraging thinking about edge cases where it doesn't hold -- I think a plain directed graph should be sufficient for supporting browseable traversal

m-mohr: I agree with @jisantuc here.

m-mohr:
Would it make sense to split the Children conformance class and the rest of the PR into two PRs? This PR is so large that it's nearly impossible for me to really do a full PR in one go...

ogcapi-features/README.md
```diff
@@ -110,6 +112,85 @@ STAC API.
...
+fields are described in detail in the [Item Search](../item-search/README.md#paging) spec.  To avoid returning
+the entire original request in the response, the  `merge` 
+property can be specified. This indicates that the client must send the same request headers that were sent in
+the original 
+request, but with the specified headers values merged in. This allows servers to indicate what needs to change 
+to get to the next page without mirroring the entire request back to the client. 
+
+See the [paging examples](../item-search/examples.md#paging-examples) for additional insight.
+
+## Collection Pagination
```
geospatial-jeff, line 155: This section is mentioned in the children conformance class (collections + catalogs) while this section is just pagination for collections, which is confusing to the reader.  Given that pagination over collections/catalogs is identical, I think this section should include catalog pagination.

core/README.md
```diff
@@ -128,3 +180,108 @@ the [overview](../overview.md#example-landing-page) document.
 ## Extensions
 
 None.
+
+## Browseable Catalogs
```
geospatial-jeff, line 184: This is a really good description.  I was having trouble groking the concept of "browseable" vs. "searchable" in the various issues, but this makes it really clear.

children/README.md
```diff
@@ -0,0 +1,156 @@
...
+  - [Endpoints](#endpoints)
+  - [Pagination](#pagination)
+  - [Example](#example)
+
+- **OpenAPI specification:** [openapi.yaml](openapi.yaml) ([rendered version](https://api.stacspec.org/v1.0.0-beta.5/children))
+- **Conformance URI:** <https://api.stacspec.org/v1.0.0-beta.5/children>
+- **Extension [Maturity Classification](../extensions.md#extension-maturity):** Pilot
+- **Dependencies**: [STAC API - Core](../core)
+
+A STAC API can return information about all STAC [Catalogs](../stac-spec/catalog-spec/catalog-spec.md) available using a link
```
geospatial-jeff, line 14: I'm confused if the server is expected to return just the first generation of children, the last generation of children, or the entire ancestry.  For example, if I had Landsat catalogs organized as `/catalogs/landsat_8_c1/{path}_{row}_{date}`, the `/children` endpoint could:
1. Return all `{path}` catalogs (immediate children).
2. Return every combination of `{path}_{row}_{date}` (last generation of children).
3. Return every combination of `{path}`, `{path}_{row}`, and `{path}_{row}_{date}` (entire ancestry).

core/README.md
```diff
@@ -128,3 +180,108 @@ the [overview](../overview.md#example-landing-page) document.
...
+        - /catalogs/landsat_8_c1/139_045_20170305
+          - /collection/landsat_8_c1/items/LC08_L1TP_139045_20170305_20170317_01_T1
+      - /catalogs/landsat_8_c1/139_046
+        - /catalogs/landsat_8_c1/139_046_20170304
+          - /collection/landsat_8_c1/items/LC08_L1TP_139046_20170304_20170316_01_T1
+        - /catalogs/landsat_8_c1/139_046_20170305
+          - /collection/landsat_8_c1/items/LC08_L1TP_139046_20170305_20170317_01_T1
+
+If done in a consistent manner, these can also provide "templated" URIs, such that a user could directly request a
+specific path, row, and date simply by replacing the values in `/catalogs/landsat_8_c1/{path}_{row}_{date}`.
```
geospatial-jeff, line 255: Is there a way for the server to specify the templated URI for exposing children catalogs?  I'm thinking of a catalog-level field containing the appropriate URI template.  For example, the `/catalogs/landsat_8_c1` catalog would have a children URI template of `/catalogs/landsat_8_c1/{path}_{row}_{date}` while the `/catalogs/landsat_8_c1/139` catalog would have a children URI template of `/catalogs/landsat_8_c1/139_{row}_{date}`.  The base of the URI template is just the `self` link to the current catalog, while the latter half is the actual template.

This would be particularly powerful when combined with a collection-level summary defining all the possible values for these templates.

children/README.md
```diff
@@ -0,0 +1,156 @@
...
+from the landing page that uses the link relation `children`, which links to an endpoint called
+`/children`. The purpose of this endpoint is to present a single resource from which clients can retrieve
+all the child objects (Catalogs and Collections) of a Catalog. This eliminates then need for a client to
```
geospatial-jeff, line 17: I think that returning both Catalogs and Collections in the `/children` endpoint is confusing.  My understanding after reading this PR is that collections are for searching while catalogs are for browsing.  It says in the `Browseable Catalogs` best practices that collections should not be included as part of the browseable tree of catalogs.  Allowing collections to be used for browsing (even though its not recommended) is poor separation of concerns.

children/README.md
```diff
@@ -0,0 +1,156 @@
+# STAC API - Children
```
geospatial-jeff, line 1: Are catalogs returned by this endpoint allowed to implement item search (i.e. can they have a `/search` endpoint)?

geospatial-jeff:
The top level README (https://github.com/philvarner/stac-api-spec/tree/catalogs-and-browseable#in-this-repository) should be updated to link to the `children` folder w/ description.

ogcapi-features/README.md
matthewhanson: Moreover, paging in general is the same across catalogs, collections, and items is it not?

children/README.md
m-mohr: Yeah, this is a bit confusing. Like do I not need to retrieve /collections anymore if /children is implemented? How should clients handle all the duplication to show a list of unique catalogs/collections?

ogcapi-features/README.md
m-mohr: I hope it is! Otherwise, we should unify it.

core/README.md
matthewhanson: +1 on this.  Also from an implementation side, where this is probably going to be generated dynamically, it would be useful for the catalog or collection to have that template specified within it so the server can use that to create the browse links dynamically.  As it stands now this would have to be specified outside of STAC...i.e., some implementation specific way

children/README.md
m-mohr: Hmm... is this a conformance class a "simple" way for providers to express their "preferred list of catalogs and collections to show on a frontpage"?

children/README.md
lossyrob: I can see the benefit of having `children` return both Catalogs and Collections - it more fits with the STAC core specification which uses `rel=child` for both of those cases. It also allows for all of the catalogs and collections to be retrieved in a single paginated call - useful for rendering the content of the API.

One addition we could make is that a query parameter can determine which type to return - e.g.

```
https://stac.api/children?type=catalog
```

children/README.md
matthewhanson: Yes, any (sub) Catalog can implement a search endpoint.

m-mohr:
> * Catalogs in a STAC API should not contain Items (as a best practice)
> * Collections in a STAC API should not contain any children (as a best practice)

I was confused by this in the call and did not pick this up as being agreed on yet. What's the background on this?

core/README.md
```diff
@@ -128,3 +180,108 @@ the [overview](../overview.md#example-landing-page) document.
...
+- / root catalog
+  - child -> /catalogs/sar
+    - child -> /catalogs/sentinel_1_l2a
+    - child -> /catalogs/afrisar
+  - child -> /catalogs/eo_boa
+    - child -> /catalogs/landsat_8_sr
+    - child -> /catalogs/sentinel_2_l2a
+
+Since the catalogs structure is a directed acyclic graph which allows 
+you to provide numerous different Catalog and Collection graphs reach leaf Items. For example, for a Landsat 8 data
```
cholmes, line 280: It might be good to provide some guidance on the final 'item' link from different views. Like should they all link to the exact same item url? Or have different item urls that all have canonical hrefs to the same one? And should the canonical one be in the browse hierarchy, or in the collection?

cholmes (commented):
Not quite done with full review, but want to get some of my comments in as not sure when I'll get back to it.

core/README.md
geospatial-jeff: OGC API Tiles has the convention of a `templated` link which could work here - https://github.com/opengeospatial/ogcapi-tiles/blob/336a40bb611b62dc9f3ee9954663bf1964bb903b/core/standard/requirements/tileset/REQ_tileset-description.adoc.

But I think this doesn't have to be a link.

children/README.md
geospatial-jeff: It's included as part of https://github.com/philvarner/stac-api-spec/blob/catalogs-and-browseable/core/README.md#browseable-catalogs introduced in this PR.  The exact language is:

> These are the two standard ways of structuring a browseable tree of catalogs, the only difference being whether the Collection is used as part of the tree or not:
> - Catalog (root) -> Catalog* -> Item (recommended)
> - Catalog (root) -> Collection -> Catalog* -> Item

core/README.md
m-mohr: I'd just like to avoid that we put something into links that existing clients don't understand ;-)

lossyrob:
@m-mohr 
> Catalogs in a STAC API should not contain Items (as a best practice)
> Collections in a STAC API should not contain any children (as a best practice)

I see this is in direct conflict with [this declaration](https://github.com/radiantearth/stac-api-spec/pull/229#discussion_r758735885) of Catalogs as a browsable unit that contains Items. When we were discussing it, I was less thinking about the "browseable" part of APIs, which to me is less interesting. The children extension is interesting for my use case in that it allows the ability to create sub-STAC APIs to organize Collections into a multi-level hierarchy, which solves an issue when you have many Collections in an API, some of them strongly related. If we consider Collections as the "searchable" container vs child Catalogs being "sub-APIs", while a Catalog that is a child of a Landing Page root of a STAC API _could_ have Items, it is confusing to have Items in Catalogs, which can't be searched, and confusing for Collections to contain anything but Items. This is true if we only want to use Catalogs in an API them as an organizational mechanism for Collections. 

Collections that contain Catalogs that organize the Items into things like date/path/row would allow for users to click through an organized set of items, enabling the "Browsable" part of all this work - I see that now. This compounds the complexity of using Catalogs for both organizing Collections (and existing as sub-APIs) and also organizing Items (which may or may not be a sub-API, though I'm not sure you'd want that many sub-APIs for groups of Items in the same Collection, though in order to have their own `/children` endpoint they may need to be a STAC API themselves). So let me retract those points.

To think through something specific and imagine an API that might have both cases:
- A root STAC API Landing Page/Catalog (which I'll refer to as a STAC) that contains a large number of Collections. These Collections are organized into Catalogs that themselves represent STACs (sub-STACs).
- The root `/collection` endpoint of the root STAC returns either _all_ collections, recursive through the tree of sub-STACs, or only the collections that have the root STAC as a parent (not sure what's best here).
- The root `/children` endpoint returns all direct children of the STAC, i.e. any Catalog (sub-STAC) or Collection that has the root STAC set as the parent
- The root `/search` endpoint will search through all collections contained in the STAC or sub-STAC, recursively.
- Say you have a sub-STAC called "MODIS" that contains all Collections for products related to modis. This is retrievable through the `/children` endpoint. It has its self link contained to the STAC endpoint which returns the Landing Page/Catalog for that sub-STAC. Say it's at `/children/modis`.
- The `/children/modis/collections` endpoint will return all the MODIS Collections, who have their `parent` link set to the MODIS sub-STAC endpoint
- The `/children/modis/children` endpoint returns all children of that sub-STAC. In this case, the MODIS sub-stack only contains Collections (and no sub-STACs of its own), so it returns the same as the `/children/modis/collections` endpoint
- The `/children/modis/search` endpoint searches against only the MODIS Collections contained by this sub-STAC
- Now let's consider a specific MODIS Collection, say `/children/modis/collections/MOD14A2`. If I call `/children/modis/collections/MOD14A2/children` ... this is where I'm confused. The Collection isn't a STAC API, so it wouldn't have a children endpoint to 'crawl' through. So it seems like the `/children/modis/children` would have to contain the Collection's subcatalogs, even though the parent of those Catalogs is actually the Collection?

Perhaps someone can help clear that up for me by continuing that specific example? I was trying to get to the point where the MOD14A2 Collection has subcatalogs that organize the Items into browsable categories, let's say year month day. Then thinking through how that would translate into a front-end experience like STAC Browser, i.e. what endpoints the front end would call when.

ogcapi-features/README.md
philvarner: All of the paginations are the same. This section shouldn't mention catalog pagination because it's specifically talking about pagination over the `/collections` endpoint from OAFeat.

This is also outdated now, as it was included b/c I'd branched this PR from another that had the pagination changes in it, which have since been merged.

I do agree about the general point that we have the same pagination description in 5 different places, and it would be nice to consolidate that. One option would be to define the pagination scheme in Core (even through it wouldn't be used) and then reference it in all of the other places.  Filed https://github.com/radiantearth/stac-api-spec/issues/236

children/README.md
philvarner: Thanks for bringing this up. I thought I'd explicitly mentioned this, but I can't find it, so I'll make sure to add something.

README.md
```diff
@@ -64,6 +64,9 @@ The *[ogcapi-features](ogcapi-features)* folder describes how a STAC API can ful
 Features](http://docs.opengeospatial.org/is/17-069r3/17-069r3.html) to expose individual `items` endpoints for search of
 each STAC collection. It also includes extensions that can be used to further enhance OAFeat.
```
philvarner, line 65: Need to add STAC API - Children here

philvarner:
I extracted a bunch of the typo and wordsmithing changes into https://github.com/radiantearth/stac-api-spec/pull/234/files

core/README.md
philvarner: updated the language to only be `directed graph`

PRINCIPLES.md
```diff
@@ -9,12 +9,17 @@ core geospatial standards.
 The collaboration facilities of Github should be used to their full extent. All proposed improvements and
 changes should come in the form of pull requests, using code review functionality to discuss changes.
 
+- **Alignment with OGC standards** - The Open Geospatial Consortium publishes a large set of geospatial standards.
```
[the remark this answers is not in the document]
philvarner: Moved this to https://github.com/radiantearth/stac-api-spec/pull/237 (I can't seem to add Jeff as a reviewer for some reason)

children/README.md
philvarner: To answer @geospatial-jeff 's question, it's up to the implementer. I think typically it will return exactly the same set of children that have `child` relations from the root. The benefit to having it at this endpoint is that the title, desc, etc can all be returned, so a client doesn't have to retrieve every single child link just to find out the title to display -- this was Rob's use case.

Maybe "simple" isn't the right word. I'll think about this.

philvarner:
One other issue to consider is that OGC uses the term "crawlable", which I think is synonymous with our use of "browseable", so we should consider adopting that term instead.

children/README.md
philvarner: I'll see if I explicitly stated it anywhere, but the intention was that `/children` would return the same list of entities that are linked to via `rel=child` from the root. The benefit (as Rob has articulated before) is that a client can get all of the entities with one call instead of having to make one HTTP request for each one to, say, only get the title of the collection or catalog.

philvarner:
> The top level README (https://github.com/philvarner/stac-api-spec/tree/catalogs-and-browseable#in-this-repository) should be updated to link to the `children` folder w/ description.

Thanks @geospatial-jeff -- I noticed the same and added it

core/README.md
```diff
@@ -128,3 +180,108 @@ the [overview](../overview.md#example-landing-page) document.
...
+product, you may want to allow browsing both by date then path then row, or by path then row then date:
+
+1. Catalog -> Catalog (product) -> Catalog (date) -> Catalog (path) -> Catalog (row)
+2. Catalog -> Catalog (product) -> Catalog (path) -> Catalog (row) -> Catalog (date)
+
+There are many options for how to structure these catalog graphs, so it will take some analysis work to figure out
```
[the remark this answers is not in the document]
philvarner: I filed an issue to come back to these https://github.com/radiantearth/stac-api-spec/issues/243 

I think this is an improvement over what we had before, but recognize there's still a better way to describe these that I can't quite articulate right now

core/README.md
philvarner: added

browseable/README.md
```diff
@@ -0,0 +1,101 @@
...
+This conformance class also requires for the endpoints in the [STAC API - Core](../core) conformance class to be implemented.
+
+This conformance class adds no additional endpoints.
+
+## Example Landing Page for STAC API - Browseable
+
+This JSON is what would be expected from an API that implements `STAC API - Browseable`. 
+
+This particular catalog provides both the ability to browse down to child Catalog objects through its
+`child` links, which then will eventually reach Items through `item` link relations.
```
cholmes, line 51: The example below doesn't seem particularly meaningful. Could be good to try to illustrate the point more, since usually the example helps show that. Perhaps a little diagram, that shows a non-browsable catalog vs a browsable one? Or perhaps just explain after the example that a catalog not implementing browsable would not have the 'child' links, but it would have them in the 'data' rel link. I think it'd be good to just provide more context, and to help make it clear to existing implementations what they need to do.

children/README.md
```diff
@@ -0,0 +1,214 @@
...
+
+The `/children` endpoint supports a pagination mechanism that aligns with pagination as described in the 
+OGC API - Common - Part 2: Geospatial Data specification. This is described in detail in
+the [STAC - Features Collection Pagination section](../ogcapi-features/README.md#collection-pagination).
+To the greatest extent possible, the catalog should be structured such that all children can be
+retrieved from the endpoint in a single call.
+
+## Example
+
+Below is a minimal example, but captures the essence. Each object in the `children` array
```
cholmes, line 67: nice example!

core/README.md
```diff
@@ -71,16 +95,27 @@ supported by this endpoint, e.g., `text/html`.
...
+search over only a sub-catalog. This is useful for very large or federated catalogs that cannot support searching
+over the entire catalog, but can support searching over individual sub-catalogs within it.
 
-It is also valid to have `item` links from the landing page, but most STAC API services are used to 
-serve up a large number of features, so they typically
-use several layers of intermediate `child` links before getting to Item objects.  Note that the `items` (plural)
-link will be used by APIs implementing STAC API - Features to link from a Collection to the items in that collection.
+Note that there is a different link relation `items` (plural)
+used by APIs conforming to the `STAC API - Features` class that links from a Collection to the items in
+that collection.
```
cholmes, line 118: Same clarification as above:

```suggestion
that collection (as a queryable paging response from the `items/` [endpoint of OGC API - Features](https://docs.opengeospatial.org/is/17-069r3/17-069r3.html#_items_)).
```

core/README.md
```diff
@@ -71,16 +95,27 @@ supported by this endpoint, e.g., `text/html`.
...
+`child` link relations may point to another STAC Catalog that *also* acts as a STAC API root, which can support
+search over only a sub-catalog. This is useful for very large or federated catalogs that cannot support searching
+over the entire catalog, but can support searching over individual sub-catalogs within it.
 
-It is also valid to have `item` links from the landing page, but most STAC API services are used to 
-serve up a large number of features, so they typically
-use several layers of intermediate `child` links before getting to Item objects.  Note that the `items` (plural)
-link will be used by APIs implementing STAC API - Features to link from a Collection to the items in that collection.
+Note that there is a different link relation `items` (plural)
+used by APIs conforming to the `STAC API - Features` class that links from a Collection to the items in
```
cholmes, line 117:
```suggestion
used by APIs conforming to the [`STAC API - Features`](../../ogcapi-features/README.md) class that links from a Collection to the items in
```

(not sure if this will render right, so may need tweaks, but just think it'd be nice to provide a link for people)

core/README.md
```diff
@@ -156,3 +204,113 @@ the [overview](../overview.md#example-landing-page) document.
 ## Extensions
 
 None.
+
+## Structuring Catalog Hierarchies
+
+A STAC API is more useful when it presents a complete `Catalog` representation of all the data contained in the
+API, such that all `Item` objects can be reached by transitively traversing `child` and `item` link relations from
+the root. This property of being able to reach all Items in this way is formalized in the
+`STAC API - Browseable` conformance class, but any Catalog can be structured for hierarchical traversal.
```
cholmes, line 213:
```suggestion
[`STAC API - Browseable` conformance class](../../browseable/README.md], but any Catalog can be structured for hierarchical traversal. 
```

core/README.md
```diff
@@ -156,3 +204,113 @@ the [overview](../overview.md#example-landing-page) document.
...
+the root. This property of being able to reach all Items in this way is formalized in the
+`STAC API - Browseable` conformance class, but any Catalog can be structured for hierarchical traversal. 
+Implementers who have search as their primary use case should consider also implementing this
+alternate view over the data by presenting it as a directed graph of catalogs, where the `child` link relations typically
+form a tree, and where each catalog can be retrieved with a single request (e.g., each Catalog JSON is small enough that
+it does not require pagination).
+
+While OAFeat requires that all Items must be part of a Collection, this does not mean that the Collection needs to be
+part of the browseable tree. If they are part of the tree, it is recommended that there only be one Collection in a
+path through the tree, and that sub-collections not be used.
```
cholmes, line 221: I'm not sure what sub-collections means here? May be worth fleshing out some.

cholmes (approved):
Looks great! I added a few commitable suggestions, but feel free to tweak them. And one or two other suggestions that are not 'must have', so I'm approving this.

I was wondering why you ended up with 'browseable', as I thought you were leaning towards 'crawlable' like OGC. I'm happy with it as is, just curious.

browseable/README.md
```diff
@@ -0,0 +1,101 @@
...
+- **Conformance URIs:** 
+  - <https://api.stacspec.org/v1.0.0-beta.5/browseable>
+  - <https://api.stacspec.org/v1.0.0-beta.5/core>
+- **Extension [Maturity Classification](../extensions.md#extension-maturity):** Pilot
+- **Dependencies**: [STAC API - Core](../core)
+
+A STAC API conforming to the `STAC API - Browseable` conformance class must be structured such that all 
+all Items in the catalog can be accessed by following `child` and `item` link relations. This is a more significant
+constraint than a STAC API without this conformance class or a STAC Catalog that is available over HTTP but does not
+implement STAC API, neither of which have any guarantee regarding the reachability of Items.
```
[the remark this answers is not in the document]
matthewhanson: +1 on @cholmes suggestion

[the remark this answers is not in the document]
matthewhanson: I can't remember when we discussed this before - did we decide that items are reachable through a single path only?

children/README.md
matthewhanson: So if an API contains the 'children' conformance class, does this mean that any sub-catalog within the tree have a 'children' link? I think I'm hung up on "shall contain" verbiage....I looked it up and in terms of a legal definition it's stated to be ambiguous as it can mean "may, will or must."

matthewhanson (approved):

matthewhanson:
This looks great, approved, I just left a comment regarding the use of the word "shall".

core/README.md
```diff
@@ -156,3 +204,113 @@ the [overview](../overview.md#example-landing-page) document.
...
+      - /catalogs/landsat_8_c1/139_045
+        - /catalogs/landsat_8_c1/139_045_20170304
+          - /collection/landsat_8_c1/items/LC08_L1TP_139045_20170304_20170316_01_T1
+        - /catalogs/landsat_8_c1/139_045_20170305
+          - /collection/landsat_8_c1/items/LC08_L1TP_139045_20170305_20170317_01_T1
+      - /catalogs/landsat_8_c1/139_046
+        - /catalogs/landsat_8_c1/139_046_20170304
+          - /collection/landsat_8_c1/items/LC08_L1TP_139046_20170304_20170316_01_T1
+        - /catalogs/landsat_8_c1/139_046_20170305
+          - /collection/landsat_8_c1/items/LC08_L1TP_139046_20170305_20170317_01_T1
```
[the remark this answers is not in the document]
lossyrob: Are you avoiding adding the child catalog links to the Collection to avoid URLs like `/collections/landsat_8_c1/139_046_20170304`, which would be hard for servers to handle cases like say a sub-Catalog named "items"? The description `Catalog (root) -> Collection -> Catalog* -> Item` lends itself to the root Catalog having children that would return the Collection, which in turn has children Catalogs*, which contain Items. I don't think anything blocks that from being a proper layout, so why duplicate a Catalog and Collection both with ID "landsat_8_c1"?

core/README.md
```diff
@@ -101,13 +136,26 @@ advertised. All service descriptions provided as part of the STAC API spec use O
 easily be used to return either YAML or JSON from this endpoint. OAFeat does not currently define a conformance
 class for OpenAPI 3.1, but may in the future. 
 
+If sub-catalogs are used, it is **recommended** that these use the endpoint `/catalogs/{catalogId}` to avoid conflicting
+with other endpoints from the root.
+
+| Endpoint              | Returns                                        | Description          |
+| --------------------- | ---------------------------------------------- | -------------------- |
+| `/catalogs/{catalogId}` | [Catalog](../stac-spec/catalog-spec/README.md) | child Catalog object |
```
lossyrob, line 144: My previous question on the example using `/catalogs/landsats-8-c1` is answered here. I see that it's a recommendation to avoid conflicts, though I think implementations could work around that and it would provide a cleaner set of URLs to avoid both a Catalog and Collection with the same name. Actually I was surprised to see that Catalog.id and Collection.id don't at least recommend there isn't a conflict inside the same root, as having duplicate IDs for Collections and Catalogs might get confusing.

lossyrob (approved):
I have some comments related to using `catalogs/` instead of `collections/` for hosting sub-catalogs for Item groupings for the Hierarchy recommendations. However I don't think these comments are blocking.

💯 well done!

geospatial-jeff (approved):
Nice work @philvarner!

browseable/README.md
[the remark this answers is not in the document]
philvarner: @matthewhanson do you mean reachable through a single path through the child/item link relations, or that they're only accessible through a single URI path (e.g., /collections/{collection_id}/items/{item_id})?

browseable/README.md
```diff
@@ -0,0 +1,101 @@
...
+every Item in the Catalog can be accessed by traversing these relations.
+
+| **rel** | **href** | **From**  | **Description**                        |
+| ------- | -------- | --------- | -------------------------------------- |
+| `child` | various  | STAC Core | The child STAC Catalogs & Collections. |
+| `item`  | various  | STAC Core | The child STAC Items.                  |
+
+Note that there is a different link relation `items` (plural)
+used by the `STAC API - Features` conformance class that links from a collection to the items in
+that collection.
```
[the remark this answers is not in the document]
philvarner: Clarified it with different language than this suggestion.

children/README.md
philvarner: I can't find the exact link right now, but OGC follows another standards body's definitions (I think ISO) where shall and must mean the same thing (you have to do this) and "may" means it's not prohibited but you don't have to. 

sub-catalogs within the tree that are also STAC APIs don't have to implement children -- each sub-catalog/sub-API is independent of the parent's conformance.

core/README.md
philvarner: I think that's a pretty reasonable recommendation to make. My only hesitancy in making it is that I feel like Catalog and Collections (as we currently define them) are not related in a way in which there would ever be confusion between them. If we'd defined a Collection as-a Catalog (maybe with the additional restriction that only one Collection can exist in a hierarchy?), then I could see a good case for not duplicating ids.

core/README.md
philvarner: I removed that word here and one other place and explicitly used "collection with a child collection"

philvarner:
> I was wondering why you ended up with 'browseable', as I thought you were leaning towards 'crawlable' like OGC. I'm happy with it as is, just curious.

My understanding was that the semantics of this may be slightly different than whatever crawlable ends up meaning in OGC, and that crawlable might not even be the term they end up using. Given that uncertainty, if we go with our own terminology, we can easily align it to theirs when then work is finalized, whereas it would be very confusing if we had something that was named the same but different.

Issue #233, opened 2021-12-03
Filter Extension: Add Accent and Case-Insensitive comparison operator conformance class(es)


philvarner:
Currently WIP in OAF, but this is expected to be the syntax https://github.com/opengeospatial/ogcapi-features/pull/641/files

PR #235, opened 2021-12-06
Filter Extension - add Accent and Case-insensitive Comparison conform…

philvarner:
**Related Issue(s):** #233

**Proposed Changes:**

1. Add mention of Accent and Case-insensitive Comparison conformance class to Filter Extension
2. add ACCENTI and CASEI examples

CQL2 spec merge is here: https://github.com/opengeospatial/ogcapi-features/pull/641/files

**PR Checklist:**


fragments/filter/README.md
```diff
@@ -1025,13 +1040,50 @@ filter=CASEI(provider) == 'NASA'
...
+### Example 13: Using the ACCENTI Accent-insensitive Comparison Function
+
+The predefined function `ACCENTI` allows for accent-insensitive comparisons. This function is
+defined in the Accent and Case-insensitive Comparison conformance class. In the example below,
+`ACCENTI('tiburon')` and `ACCENTI('tiburón')` evaluate to be equal.
+
+#### Example 13: cql2-text (GET)
+
+```http
+filter=ACCENTI(provider) = ACCENTI('tiburón')
```
m-mohr, line 1067: What's the reasoning to not include the ACCENTI functionality into CASEI?
From a German language perspective this is *really* weird as "Straße" works in CASEI, but "Süße"  (I assume) does not and I have to use ACCENTI (while ä does not contain an accent for us). ß and ü are both special characters in German, but are on the same level. It's hard to distinguish which to use when, especiialy as ü is not called an accent (like ú or so).

m-mohr: Ah, found it: https://github.com/opengeospatial/ogcapi-features/pull/641#issuecomment-987853664

m-mohr (commented):
Looks good, but maybe don't highlight/show ACCENTI in STAC API yet until it has been decided on?

fragments/filter/README.md
philvarner: I think two points here:
1) My understanding is that the unicode rules for case include ss/ß even though it's not really the same as other case changes like a/A. 
2) Accent here is used in a very general sense, rather than a precise technical one. I think technically, only acute and grave are accents, and circumflex and umlauts (and every other way to "modify" ASCII characters) fall into the larger category of diacritics.

philvarner:
> Looks good, but maybe don't highlight/show ACCENTI in STAC API yet until it has been decided on?

Yeah, I think I'm going to put this PR in draft and wait until this gets nailed down a bit more.

philvarner:
Accent and Case insensitive comparison class code was merged to the CQL2 spec this morning, so this pr is ready.

matthewhanson (approved):

PR #253, opened 2022-01-13
Filterfragmentfixes

bitner:
**Related Issue(s):** #

**Proposed Changes:**

1. Fix examples in fragments/filter/README.md so that cql2-text examples are exactly equivalent to cql2-json examples. (This allows use within tests to make sure that a cql2-text parser returns the same AST as a cql2-json parser.)

**PR Checklist:**


bitner:
@philvarner I added some fixes to the examples so that the examples can be scraped and used to make sure that the pygeofilter cql2-text and cql2-json parsers have identical results.

Issue #241, opened 2021-12-11
Update Filter OpenAPI doc to match current CQL2 spec


philvarner:
Deferring this to rc.1, to allow the CQL2 spec to stabilize more.

PR #261, opened 2022-02-28
update CQL2 definition to reference OGC API yaml file in github, update npm build

philvarner:
**Related Issue(s):** 

- #241

**Proposed Changes:**

1. Update CQL2 spec to reference version in ogcapi-features github repo 
2. npm build improvements

**PR Checklist:**


.gitignore
```diff
@@ -35,7 +35,6 @@ build/Release
 # Dependency directories
 node_modules/
 jspm_packages/
-package-lock.json
```
philvarner, line 38: recommended practice is to commit package-lock.json

.nvmrc
```diff
@@ -0,0 +1 @@
+16
```
philvarner, line 1: use nvmrc, and upgrade to node 16

collections/openapi.yaml
```diff
@@ -12,14 +12,15 @@ info:
     name: Apache License 2.0
     url: 'http://www.apache.org/licenses/LICENSE-2.0'
 tags:
+  - name: Core
```
philvarner, line 15: fixes a few warnings.

fragments/filter/README.md
```diff
@@ -222,8 +222,6 @@ Formal definitions and grammars for CQL2 can be found here:
   - [cql.bnf](https://github.com/opengeospatial/ogcapi-features/blob/master/extensions/cql/standard/schema/cql.bnf)
   - [cql.json](https://github.com/opengeospatial/ogcapi-features/blob/master/extensions/cql/standard/schema/cql.json)
   - [cql.yml](https://github.com/opengeospatial/ogcapi-features/blob/master/extensions/cql/standard/schema/cql.yml)
-- A JSON Schema for only the parts of the CQL2 JSON encoding required by this extension is [here](cql.json)
```
philvarner, line 225: edited cql.json has been removed, as we no longer have a custom version

cholmes (approved):
This looks good, but seems to also include some various clean-up updates in the config, pr template, gitignore, etc. Ideally those would be a separate PR. I don't think that's necessary, but maybe also mention those in the title of the PR?

.gitignore
m-mohr: For software yes, but for a specification? There I'd assume you'd want the latest version?

.nvmrc
m-mohr: What's the reasoning? And why in a CQL PR?

philvarner: I was updating the openapi yaml, so I took the opportunity to update the build which mostly does validation while I was at it.  I probably should have separated them into different PRs, but didn't.

Issue #242, opened 2021-12-14
Is Query Extension (STACQL) still going to be deprecated?


philvarner:
Needs discussion.

matthewhanson:
Discussed in working group
No need to deprecate, but instead organize into separate repos/versions (see #144 , #84)

PR #262, opened 2022-02-28
Undeprecate query extension

philvarner:
**Related Issue(s):** 

- #242 

**Proposed Changes:**

1. remove deprecated language for Query Extension

**PR Checklist:**


cholmes (approved):

m-mohr:
Shouldn't it move to a separate repo instead?

philvarner:
> Shouldn't it move to a separate repo instead?

I think we should move some (if not all) of them to a separate repo, but this is just to update it so it's accurate.

Issue #251, opened 2022-01-10
Require that /children must reflect the root children?


m-mohr:
The Children conformance class says (highlighted text by me):

> It is **recommended** that the Catalog and Collection objects returned by the /children endpoint are the same as those referenced by child link relations in the STAC API Landing Page.

Why is this only recommended? For client implementations, it's much easier to implement if the links in the root landing page child links are "equal" to the /children response. Why? Let me give a short example:

STAC Browser requests/has:
- /collections (which may or may not contain what the others have)
- /children (which is allowed to contain more or fewer children than the landing page)
- landing page child and item links (which is allowed to contain more or fewer children than /children)

Now I need to merge all three sources together **and check for duplicates**, which is a pain and error-prone as some may only provide a URL and some may only provide an ID. Can we improve this situation somehow?

This get's especially annoying if I try to paginate / any or multiple of the sources have pagination in place. Like how shall I paginate something from three different sources at the same time?

matthewhanson:
Discussed inn working session.

- children conformance class needs to require that /children includes all child links

PR #265, opened 2022-02-28
require children endpoint to return all child link objects

philvarner:
**Related Issue(s):** 

- #251

**Proposed Changes:**

1. require /children to return all child links

**PR Checklist:**


cholmes (approved):

children/README.md
```diff
@@ -14,16 +14,15 @@
...
-to present a single resource from which clients can retrieve
-all the immediate child objects of a Catalog, which may be Catalog or Collection objects.
-The `child` link relations in a Catalog already allow for describing these
-relationships, but require a client to retrieve each resource URL to find any information about the children
-(e.g., title, description), which can cause significant performance issues in user-facing applications.
-
-It is recommended that the Catalog and Collection objects returned by the `/children` endpoint are the same
-as those referenced by `child` link relations in the STAC API Landing Page. Following these semantics, it should
-return the immediate children of the root Catalog, rather than all descendant catalogs or collections.
+it contains using the link relation `children` to an endpoint `/children`. The `/children` endpoint shall
```
m-mohr, line 17: Just a general observation, but I think we mostly use MUST instead of SHALL in the documentation. Your recent PRs use SHALL and I think we should be more consistent here. Could we try to use MUST instead of SHALL? I just had to look it up again whether SHALL was equal to MUST or SHOULD and found it confusing.

philvarner: I just looked. We use must a lot more than shall. I'd prefer must also, since shall seems overly formal in English. I'll file an issue to make these consistent.

PR #266, opened 2022-02-28
add type=application/geo+json to prev and next links

philvarner:
**Related Issue(s):** #256

**Proposed Changes:**

1. add type=application/geo+json to prev and next link relations

**PR Checklist:**

### DIFF
--- a/item-search/README.md
+++ b/item-search/README.md
@@ -152,11 +152,13 @@ parameter name is defined by the implementor and is not necessarily part of the 
     "links": [
         {
             "rel": "next",
-            "href": "http://api.cool-sat.com/search?page=3"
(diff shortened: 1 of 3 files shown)


m-mohr:
Just realized a day ago, but is it in a conflict that the type is geo+json instead of just JSON and errors being given as plain json?

philvarner:
It's not a conflict. This is more just an advertisement that (1) if you sent an `Accept` header with this value (or no Accept header) it will respond with a 200 and that type as the body. Any non-200 response isn't expected to return the same type.

cholmes (approved):

m-mohr (approved):

Issue #255, opened 2022-02-16
Clarify landing page requirement for "search"


aaime:
The current specification for the item-search link relation is here:

https://github.com/radiantearth/stac-api-spec/tree/master/item-search#link-relations

One can read it in two different ways:

1) There must be only one link from landing page to search resource, must have a "search" rel and a type of "application/geo+json"
2) There must be at least one link from the landing page to search resource  that has a "search" rel and a type of "application/geo+json"

Interpretation 1) seems to be at odds with Web API multiple representations (can't I have a HTML representation too?), but simplifies life to clients that can just pick the first link the find with a "search" rel.
Interpretation 2) would be more common in OGC APIs, but seems at odds with this wording "the search link relation shall have a type of application/geo+json"

Also, some OGC API implementations have stopped providing "type" completely and rely on HTTP content negotiation to find a representation that both clients and servers can agree on, while the spec above seems to mandate the use of "type" instead.
See for example this LDProxy landing page (not a STAC one, but check the lack of "type" in links to other resources, like "data"):
https://demo.ldproxy.net/cshapes?f=json

PR #267, opened 2022-02-28
clarify wording of search link rel requirement

philvarner:
**Related Issue(s):**

- #255

**Proposed Changes:**

1. update language around search link rel

**PR Checklist:**

### DIFF
--- a/item-search/README.md
+++ b/item-search/README.md
@@ -55,8 +55,11 @@ The following Link relations shall exist in the Landing Page (root).
 | -------- | --------- | ---------------------- | --------------------------- |
 | `search` | `/search` | STAC API - Item Search | URI for the Search endpoint |
 
-The `search` link relation shall have a `type` of `application/geo+json` and a `method` of `GET`, and may also
-a link with a `method` of `POST` if the server supports it.
+This `search` link relation shall have a `type` of `application/geo+json`. If no `method` attribute is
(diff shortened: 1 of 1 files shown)


cholmes (approved):

item-search/README.md
```diff
@@ -55,8 +55,11 @@ The following Link relations shall exist in the Landing Page (root).
 | -------- | --------- | ---------------------- | --------------------------- |
 | `search` | `/search` | STAC API - Item Search | URI for the Search endpoint |
 
-The `search` link relation shall have a `type` of `application/geo+json` and a `method` of `GET`, and may also
-a link with a `method` of `POST` if the server supports it.
+This `search` link relation shall have a `type` of `application/geo+json`. If no `method` attribute is
```
m-mohr, line 58: I think it only SHOULD have it or let's say the default for search should probably be `application/geo+json`. We do the same with most other links, e.g. child, item, ...

m-mohr: And a general observation, but I think we mostly use MUST instead of SHALL in the documentation. Your recent PRs use SHALL and I think we should be more consistent here. Could we try to use MUST instead of SHALL? I just had to look it up again whether SHALL was equal to MUST or SHOULD and found it confusing.

philvarner: Filed https://github.com/radiantearth/stac-api-spec/issues/275 to discuss

PR #271, opened 2022-03-02
Claify Transaction Extension behavior

philvarner:
**Related Issue(s):** 

- https://github.com/radiantearth/stac-api-spec/issues/285

**Proposed Changes:**

1. Clarify that PUT and PATCH cannot change the collection an item is in 
2. Clarify allowed HTTP status codes
3. Allow partial Item and partial ItemCollection containing partial Items as POST input
4. Allow partial Item containing partial Items as PUT input (but the server has to be able to derive a valid Item from it to return from the GET endpoint, e.g., by populating the id and collection fields or adding links)

**PR Checklist:**


ogcapi-features/extensions/transaction/openapi.yaml
```diff
@@ -219,11 +278,6 @@ components:
             $ref: '../../../core/commons.yaml#/components/schemas/link'
         assets:
           $ref: '../../../core/commons.yaml#/components/schemas/assets'
-      example:
```
philvarner, line 222: I think this missed being deleted when this block was copied from the Item schema

CHANGELOG.md
```diff
@@ -15,6 +15,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- Clarified behavior of Transaction Extension endpoints, notably that PUT or PATCH that changes the `collection`
```
m-mohr, line 18: It seems these points are not captured in the CHANGELOG:

> Allow partial Item and partial ItemCollection containing partial Items as POST input
> Allow partial Item containing partial Items as PUT input (but the server has to be able to derive a valid Item from it to return from the GET endpoint, e.g., by populating the id and collection fields or adding links)

ogcapi-features/extensions/transaction/README.md
```diff
@@ -24,14 +24,63 @@ OGC API [Simple Transactions](http://docs.opengeospatial.org/DRAFTS/20-002.html)
 once it is released STAC will align to a released one, but we anticipate few changes as it is a very simple document.
 
 STAC Transactions additionally support optimistic locking through use of the ETag header, as specified in the
-OpenAPI document. This is not currently specified in OGC API - Features, but it is compatible and we will 
+OpenAPI document. This is not currently specified in `OGC API - Features`, but it is compatible and we will
```
m-mohr, line 27: Code tag seems weird for a name?

philvarner: We've been doing it in other places too. The main problem is that it's unclear in a sentence that "OGC API - Features" is a proper name, but most people read it as parenthetical phrase. Maybe we should replace all of the instances of the names of conformance classes to be double-quoted rather than code quoted?

ogcapi-features/extensions/transaction/README.md
```diff
@@ -24,14 +24,63 @@ OGC API [Simple Transactions](http://docs.opengeospatial.org/DRAFTS/20-002.html)
...
-| `PATCH /collections/{collectionId}/items/{featureId}`  | `application/json`  | Updates an existing item by ID using a partial item description, compliant with [RFC 7386](https://tools.ietf.org/html/rfc7386). |
-| `DELETE /collections/{collectionID}/items/{featureId}` | n/a                 | Deletes an existing item by ID.                                                                                                  |
+| Path                                                   | Content-Type Header | Body                                   | Success Status | Description                                                       |
+| ------------------------------------------------------ | ------------------- | -------------------------------------- | -------------- | ----------------------------------------------------------------- |
+| `POST /collections/{collectionID}/items`               | `application/json`  | partial Item or partial ItemCollection | 201, 202       | Adds a new item to a collection.                                  |
```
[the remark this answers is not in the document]
philvarner: the endpoints don't require that a valid Item be submitted. Without this, the behavior of automatically populating the "collection" attribute if it was missing would not be allowed, as the invalid Item should be rejected by the endpoint. Also some other fields like `links` are mandatory, and I wouldn't expect a client to have to post an Item with a `links` attribute with an empty array value just to get the server to accept the input.

CHANGELOG.md
philvarner: I'll add them

ogcapi-features/extensions/transaction/README.md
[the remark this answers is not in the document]
m-mohr: Okay, is this compliant with how OGC defined it as well or how do they define POSTing? I assume the have the same issues (except for the collection issue).

ogcapi-features/extensions/transaction/README.md
m-mohr: Yeah, this is completely inconsistent throughout the spec. My first intuition though would be italic instead of code, which seems semantically better and probably also is better for accessibility.

Then it's

> This is not currently specified in *OGC API - Features*, but it is compatible and we will 

instead of

> This is not currently specified in `OGC API - Features`, but it is compatible and we will

philvarner: I'm going to leave this as it is and then we can do all that search and replace in one PR.

ogcapi-features/extensions/transaction/README.md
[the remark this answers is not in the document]
philvarner: I would say that can be consistent with one change. I believe the correct semantics for POST and PUT are that the JSON entity has to at least be a valid Feature, but doesn't have to be a valid Item. This implies it must have the id field populated., but doesn't have to have collection, links, bbox, maybe others. I'll update the OpenAPI to reflect that. 

I looked yesterday and they have no explicit mention of not changing the `id`, but I think that should be implied (I'll follow up with a comment about it in their repo).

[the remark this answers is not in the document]
philvarner: updated.

philvarner:
Filed:

- https://github.com/opengeospatial/ogcapi-features/issues/678

philvarner:
Filed:

- https://github.com/radiantearth/stac-api-spec/issues/274

philvarner:
@m-mohr ready for re-review

matthewhanson (approved):

PR #287, opened 2022-03-10
fix a few minor openapi spec issues, upgrade remark and spectral

philvarner:
**Related Issue(s):** n/a

**Proposed Changes:**

1. cleanup a few minor issues with openapi specs
1. upgrade remark and spectral

**PR Checklist:**


.circleci/.spectral-fragments.yml
```diff
@@ -3,6 +3,6 @@ rules:
   openapi-tags: off
   info-contact: off
   oas3-api-servers: off
-  oas3-unused-components-schema: off
+  oas3-unused-component: off
```
philvarner, line 6: the name of this changed between spectral 5 and 6

.circleci/rc.yaml
```diff
@@ -11,8 +11,6 @@ plugins:
   - remark-lint-hard-break-spaces
   - remark-lint-blockquote-indentation
   - remark-lint-no-consecutive-blank-lines
-  - - remark-lint-maximum-line-length
```
philvarner, line 14: this actually started causing warnings on table lines in spectral 6, but we don't need to change this from the default anyway, so remvoing

children/openapi.yaml
```diff
@@ -107,16 +107,6 @@ components:
         application/json:
           schema:
             $ref: '#/components/schemas/children'
-    InvalidParameter:
```
philvarner, line 110: latest spectral determined these are unused

philvarner:
We have this configured to fail on warnings now, so putting this in draft until we figure out why the max-line-lenght rule is triggering on tables

PR #291, opened 2022-03-17
v.1.0.0-rc.1 release merge to master

philvarner:


m-mohr:
I think I don't agree with this release plan. Shouldn't all major issues being resolved for an rc.1? Especially the repo split I'd consider like it should be part of rc.1 as quite a number of things could go wrong and potentially also the conformance classes (i.e. how the URLs looks like) could change drastically. Would like to discuss in a broader group (e.g. the PSC or STAC meeting), I think.

CHANGELOG.md
```diff
@@ -4,6 +4,37 @@ All notable changes to this project will be documented in this file.
...
+### Changed
+
+- Query Extension is no longer deprecated.
+- Children conformance class now requires the /children endpoint includes all child catalogs or collections
+  referenced via `child` link relations from the Landing Page
+- Clarified behavior of Transaction Extension endpoints:
+  - PUT and PATCH of a body that changes the `collection` or `id` is disallowed.
+  - POST, PUT, and PATCH do not need to include the `collection` attribute, as it should be derived from the URL.
+  - POST and PUT can be used with a body that is at least a GeoJSON Feature, but does not have to be an Item, but for which 
+    the server can derive a valid Item, e.g., by populating the id and collection fields or adding links
```
jonhealy1, line 25: does this mean the geojson feature has to have an id field?

ogcapi-features/extensions/transaction/README.md
```diff
@@ -24,14 +24,63 @@ OGC API [Simple Transactions](http://docs.opengeospatial.org/DRAFTS/20-002.html)
...
+| `POST /collections/{collectionID}/items`               | `application/json`  | partial Item or partial ItemCollection | 201, 202       | Adds a new item to a collection.                                  |
+| `PUT /collections/{collectionId}/items/{featureId}`    | `application/json`  | partial Item                           | 200, 202, 204  | Updates an existing item by ID using a complete item description. |
+| `PATCH /collections/{collectionId}/items/{featureId}`  | `application/json`  | partial Item                           | 200, 202, 204  | Updates an existing item by ID using a partial item description.  |
+| `DELETE /collections/{collectionID}/items/{featureId}` | n/a                 | n/a                                    | 200, 202, 204  | Deletes an existing item by ID.                                   |
+
+### POST
+
+When the body is a partial Item:
+
+- Shall only create a new resource.
```
jonhealy1, line 43: Should 'shall' be taken out here as well?

ogcapi-features/extensions/transaction/README.md
```diff
@@ -24,14 +24,63 @@ OGC API [Simple Transactions](http://docs.opengeospatial.org/DRAFTS/20-002.html)
...
+| ------------------------------------------------------ | ------------------- | -------------------------------------- | -------------- | ----------------------------------------------------------------- |
+| `POST /collections/{collectionID}/items`               | `application/json`  | partial Item or partial ItemCollection | 201, 202       | Adds a new item to a collection.                                  |
+| `PUT /collections/{collectionId}/items/{featureId}`    | `application/json`  | partial Item                           | 200, 202, 204  | Updates an existing item by ID using a complete item description. |
+| `PATCH /collections/{collectionId}/items/{featureId}`  | `application/json`  | partial Item                           | 200, 202, 204  | Updates an existing item by ID using a partial item description.  |
+| `DELETE /collections/{collectionID}/items/{featureId}` | n/a                 | n/a                                    | 200, 202, 204  | Deletes an existing item by ID.                                   |
+
+### POST
+
+When the body is a partial Item:
```
jonhealy1, line 41: A geojson feature would be a partial item I think?

ogcapi-features/extensions/transaction/README.md
```diff
@@ -24,14 +24,63 @@ OGC API [Simple Transactions](http://docs.opengeospatial.org/DRAFTS/20-002.html)
...
+
+- Shall only create a new resource.
+- Shall return 409 if an Item exists for the same collection and id values.
+- Shall populate the `collection` field in the Item from the URI.
+- Shall return 201 and a Location header with the URI of the newly added resource for a successful operation.
+- May return the content of the newly added resource for a successful operation.
+
+When the body is a partial ItemCollection:
```
jonhealy1, line 49: an ItemCollection that contains partial Items?

ogcapi-features/extensions/transaction/README.md
```diff
@@ -24,14 +24,63 @@ OGC API [Simple Transactions](http://docs.opengeospatial.org/DRAFTS/20-002.html)
...
+
+- Shall only create a new resource.
+- Shall return 409 if an Item exists for any of the same collection and id values.
+- Shall populate the `collection` field in each Item from the URI.
+- Shall return 201 without a Location header.
+- May create only some of the Items in the ItemCollection.
```
jonhealy1, line 55: Maybe more explanation. When would only some Items be created?

ogcapi-features/extensions/transaction/README.md
```diff
@@ -24,14 +24,63 @@ OGC API [Simple Transactions](http://docs.opengeospatial.org/DRAFTS/20-002.html)
...
+- Shall return 202 if the operation is queued for asynchronous execution.
+- Shall return 404 if no Item exists for this resource URI.
+- If the `id` or `collection` fields are different from those in the URI, status code 400 shall be returned.
+
+PATCH is compliant with [RFC 7386](https://tools.ietf.org/html/rfc7386).
+
+### DELETE
+
+- Shall return 200 or 204 for a successful operation.
+- Shall return a 202 if the operation is queued for asynchronous execution.
```
jonhealy1, line 85: If it's queued it's not necessarily successful? Is this the best response you can give with asynchronous execution?

jonhealy1 (approved):
Looks great. I left some questions/ suggestions but I think they all reflect my limited knowledge.

ogcapi-features/extensions/transaction/README.md
philvarner: This comes from OAFeat Part 4 -- yes, it's "successfully queued" rather than actually successful, since that's the best you can do.

ogcapi-features/extensions/transaction/README.md
philvarner: It's defined in the OpenAPI spec -- it's the same as an ItemCollection, but `features` contains partial Items (e.g., don't have to have `links`, etc). The gist of the partial Item/ItemCollection is that you don't have to submit a valid Item, it just has to be enough of an Item that the server can synthesize the missing elements -- for example, `collection` can be populated from the endpoint it was posted to, only of one of bbox or geometry needs to be set, etc.

ogcapi-features/extensions/transaction/README.md
philvarner: yes, mostly. geojson features don't have to have an `id` field, so this item either has to have an id field or the server has to support autogenerating one.

ogcapi-features/extensions/transaction/README.md
philvarner: yeah, I copied this from OAFeat, so this should be must.

CHANGELOG.md
philvarner: good question -- I had thought geojson feature required id, but now I see it's a "should". So maybe we need to clarify that behavior -- maybe just say that it's up to the implementer if they want to autogenerate IDs.

philvarner:
@m-mohr that's fine, but I'm not able to do any more of this work. I've been doing this in my spare time for the last two months, and I was just trying to get an rc.1 out. If you or the PSC think any of this work should be put in for rc.1, you'll need recruit people to do it.

philvarner:
@m-mohr also, I don't think the conformance class URIs will change just because we put them in different repos. We should keep them the same, and maybe change the convention for new ones rather the make the breaking change with these.

m-mohr:
@philvarner I can totally relate to that as I've also contributed a fair amount of spare time to stac-spec work and all your work is appreciated! There's a PSC meeting on Monday so I think we can discuss this promptly.

About the conformance classes: I'm not 100% sure about that, all the URLs had changed when we moved the stac extensions. I guess the basic question is whether these URIs are "virtual" or whether there's something behind it that needs to be deployed by CI. Then it gets tricky because the CI of this repo would need to deploy stuff from external repos, right? But maybe I'm on a completely wrong track here.

cholmes:
> About the conformance classes: I'm not 100% sure about that, all the URLs had changed when we moved the stac extensions. I guess the basic question is whether these URIs are "virtual" or whether there's something behind it that needs to be deployed by CI.

The OGC precedent is very much that they are 'virtual' - they are just a bundle of functionality. Indeed the OGC stuff is annoying because they are technically URI's (I believe), not resolving to anything (like a URL does). So if you just stick an OGC conformance class URI in it often doesn't exist as a webpage at all. I'd like to do something better, but right now we're not doing anything either. The core stac-spec stuff wasn't trying to follow OGC conformance class style stuff. Those were actual JSON schemas, and we wanted to deploy them from the repos. 

Looking at OGC API features it looks like their conformance classes URI's do resolve, but they are just a redirection to the published specification. 

So I think it's fine if we move things out and then figure out how to get all the redirections right later. They deploy to api.stacspec.org, so it's not actually tied to the repository. I agree there may be some advanced CI work to make it all happen automatically. But I believe it's fine to just not have some of the conformance classes resolve - they are just a communication of the bundle of functionality, not a contract that the URL will be the openapi docs (as it is right now). Indeed I could see us choosing to redirect to the actual spec (like features API does, though they publish the spec instead of pointing direct at the github repo), or even nice documentation that explains each conformance class in more introductory texxt.

m-mohr:
Okay, I also want to ensure clients don't need to check too many conformance URIs. Right now in STAC Browser I'm doing regexp matching (like `https://api.stacspec.org/1\.[^\/]+/item-search#sort`) and I really would like to avoid that we later have to check a lot of different URIs.

m-mohr:
As long as the conformance class URIs stay the same structure then I'm good to release rc.1.

lossyrob:
Captured actionable comments in https://github.com/radiantearth/stac-api-spec/issues/292, @jonhealy1 feel free to update if I missed anything.

philvarner:
🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉

CHANGELOG.md
philvarner: Decided the answer to this is "yes". The openapi spec already defined `id` as a required field.

PR #298, opened 2022-04-28
Collections list is just plain JSON

tschaub:
**Proposed Changes:**

1. Update example link type to be `application/json` instead of `application/geo+json` for collections list.

**PR Checklist:**

- [X] A CHANGELOG entry is not required (I don't think).

### DIFF
--- a/ogcapi-features/README.md
+++ b/ogcapi-features/README.md
@@ -177,12 +177,12 @@ previous (page=2) pages:
   {
     "rel": "prev",
     "href": "https://stac-api.example.com/collections?page=2"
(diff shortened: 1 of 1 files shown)


philvarner:
CI is broken right now because of #301 , that needs to be resolved before we can merge. (it's an easy fix, someone just needs to do it)

tschaub:
#302 updates the `openapi.yaml` to use `booleanExpression`

philvarner:
updated this to be against dev instead of master

philvarner (approved):

PR #308, opened 2022-06-21
Use main instead of to dev/master

philvarner:
**Related Issue(s):** 

- #259

**Proposed Changes:**

Change branch strategy to use main w/ tags instead of dev/master. 

Former PR here: https://github.com/radiantearth/stac-api-spec/pull/268

There are two aspects to this change:
1. **Changing from using a develop->released branch model to a single branch** -- Until now, the `dev` branch was a rapidly-changing work-in-progress. It made sense to not make this the first thing people saw when they came to this repo. As we converge towards 1.0.0, there will be fewer significant changes in main that are not part of the latest released version. Most repos don't have master or a branch with their latest released version as the default branch, so users are used to `main` being the development tip.
2. **Changing the name of the primary branch from master to main** -- This has been a trend in the last couple years, motivated primarily by the origination and metaphorical use of the term "master" because of its use in the chattel slavery system in the United States prior to 1865. A project with a diverse audience should not use terms that are so blatantly offensive.

References:

- https://github.com/github/renaming
- [Regarding Git and Branch Naming](https://sfconservancy.org/news/2020/jun/23/gitbranchname/)
- [Why GitHub renamed its master branch to main](https://www.theserverside.com/feature/Why-GitHub-renamed-its-master-branch-to-main)

**PR Checklist:**

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@
 
 ## Releases (stable)
 
-- [v1.0.0-rc.1](https://github.com/radiantearth/stac-api-spec/tree/v1.0.0-rc.1)
+- [v1.0.0-rc.1](https://github.com/radiantearth/stac-api-spec/tree/v1.0.0-rc.1) (latest)
 - [v1.0.0-beta.5](https://github.com/radiantearth/stac-api-spec/tree/v1.0.0-beta.5)
 - [v1.0.0-beta.4](https://github.com/radiantearth/stac-api-spec/tree/v1.0.0-beta.4)
 - [v1.0.0-beta.3](https://github.com/radiantearth/stac-api-spec/tree/v1.0.0-beta.3)
--- a/README.md
+++ b/README.md
@@ -28,7 +28,7 @@
 The [main](https://github.com/radiantearth/stac-api-spec/tree/main) branch in GitHub is
 used for active development and may be unstable. Implementers should reference one of
 the release branches above for a stable version of the specification.
-*NOTE: This means that if you are on github.com/radiantearth/stac-api-spec then you are looking at an unreleased,
+**NOTE**: This means that if you are on github.com/radiantearth/stac-api-spec then you are looking at an unreleased,
 unstable version of the specification. Use the first listed link on releases to read the current released, stable version
 of the spec.
 ## About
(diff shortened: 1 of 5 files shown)


philvarner:
To address possible confusing about main being the development version, the top of the top-level README tools like this now:

image

philvarner:
The second part of this that @cholmes brings up is that we should really convert this spec from intending to be a set of markdown docs in github to a generated website (e.g., like tools like Sphinx do).

README.md
```diff
@@ -4,24 +4,31 @@
...
 - [v1.0.0-beta.2](https://github.com/radiantearth/stac-api-spec/tree/v1.0.0-beta.2)
 - [v1.0.0-beta.1](https://github.com/radiantearth/stac-api-spec/tree/v1.0.0-beta.1)
 - [v0.9.0](https://github.com/radiantearth/stac-api-spec/tree/v0.9.0)
 
+## Development (unstable)
+
+The [main](https://github.com/radiantearth/stac-api-spec/tree/main) branch in GitHub is
+used for active development and may be unstable. Implementers should reference one of
+the release branches above for a stable version of the specification.
+
```
cholmes, line 31: I think we should add one more note to make it more clear that if you click around from the base repo it won't be stable. Something like:

```suggestion
*NOTE: This means that if you are on github.com/radiantearth/stac-api-spec then you are looking at an unreleased,
unstable version of the specification. Use the first listed link on releases to read the current released, stable version
of the spec.
```

cholmes (approved):
Looks good, added a suggestion to be more clear in the readme, to help non-developers understand what's going on. And like you mentioned I definitely think we should move towards a 'publish' step. The new [stac website](https://github.com/radiantearth/stac-site) may offer some way to do it, but that's beyond my website & CI skills.

philvarner:
I took a brief look at eleventy (the static site generator that's being used for stacspec.org) and was able to generate a site from the spec, though it didn't have any styling and most of the links were subtly wrong. But, I think it is feasible to use this to generate websites for stac-spec and stac-api-spec that we can publish, we just need to put a bit of work into it.

matthewhanson (approved):

Issue #292, opened 2022-03-21
Transaction extension clarifications


lossyrob:
A few small changes to the transaction statement based on @jonhealy1 comments on #291

#### Clarify POST body language

https://github.com/radiantearth/stac-api-spec/pull/291#discussion_r829994090

Clarify if the posted data would need to have an ID.

### Shall -> Must for creating new resource

From https://github.com/radiantearth/stac-api-spec/pull/291#discussion_r829740770

#### Clarify partial item update

Based on this comment: https://github.com/radiantearth/stac-api-spec/pull/291#discussion_r829743967

For the line in the transaction README that reads:

> May create only some of the Items in the ItemCollection.

Expand this text to describe when only some items would be created and describe recommended behavior as per Phil's linked comment.

jonhealy1:
@lossyrob Looks great. Thanks.

PR #310, opened 2022-06-21
update wording in transaction extension

philvarner:
**Related Issue(s):** 

- #292

**Proposed Changes:**

1. update wording in transaction extension

**PR Checklist:**

### DIFF
--- a/ogcapi-features/extensions/transaction/README.md
+++ b/ogcapi-features/extensions/transaction/README.md
@@ -40,47 +40,54 @@ work to get it incorporated.
 
 When the body is a partial Item:
 
-- Shall only create a new resource.
-- Shall return 409 if an Item exists for the same collection and id values.
-- Shall populate the `collection` field in the Item from the URI.
-- Shall return 201 and a Location header with the URI of the newly added resource for a successful operation.
+- Must only create a new resource.
+- Must have an id field.
+- Must return 409 if an Item exists for the same collection and id field values.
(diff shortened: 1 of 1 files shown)


ogcapi-features/extensions/transaction/README.md
```diff
@@ -40,47 +40,54 @@ work to get it incorporated.
...
+- Each Item in the ItemCollection must have an id field.
+- Must return 409 if an Item exists for any of the same collection and id values.
+- Must populate the `collection` field in each Item from the URI.
+- Must return 201 without a Location header.
+- May create only some of the Items in the ItemCollection. Implementations are not
+  required to implement all-or-none sematics for this operation. For example, if an
+  ItemCollection contains two Items and one is successfully created and the other
+  fails to be created, the server is not required to then delete the successfully
+  created one. When only some of the Items in the ItemCollection are created, the
+  server should communicate this failure back to the client with an error status code.
```
m-mohr, line 62: Should it be defined which error code should be used when only a partial update happened?

m-mohr (approved):
Looks good to me. One issue that might be clarified in the comment above.

ogcapi-features/extensions/transaction/README.md
philvarner: I don't think we can. It could be any problem. I think the point here is that it's not atomic, and you may have partial updates, and that the server isn't required to enforce atomicity.

PR #312, opened 2022-06-27
Fix conformance classes

m-mohr:
It looks like the conformance classes were copy/pasted from item search, but instead they should contain ogcapi-features.

**PR Checklist:**


philvarner (approved):

m-mohr:
Ouch, I forgot the CHANGELOG entry here...

PR #320, opened 2022-07-26
fix stac logo image

philvarner:
**Related Issue(s):** 

- n/a

**Proposed Changes:**

1. fix stac logo image, pointing a now-non-exisant github url from the old site

**PR Checklist:**


m-mohr:
I actually added the image to stac-site again as it's broken in a couple of repos and no one will have the time to fix it everywhere...

Issue #263, opened 2022-02-28
Review Code of Conduct


philvarner:

philvarner:
1. Update to latest CCC https://www.contributor-covenant.org/version/2/1/code_of_conduct/
2. Update email address

PR #326, opened 2022-08-24
update code of conduct to 2.1

philvarner:
**Related Issue(s):** 

- #263 

**Proposed Changes:**

1. Update code of conduct to version 2.1

**PR Checklist:**

### DIFF
--- a/.remarkignore
+++ b/.remarkignore
@@ -1,1 +1,2 @@
 /CHANGELOG.md
+/CODE_OF_CONDUCT.md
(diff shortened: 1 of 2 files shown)


m-mohr:
I don't expect any objections, but I assume something like this should be approved in the PSC meeting?

cholmes (approved):

cholmes:
I think if we get majority of PSC approving PR we don't need to do it in a meeting.

matthewhanson (approved):

Issue #330, opened 2022-09-12
Incorporate Collections spec into Features spec


philvarner:
the collections conformance class is a subset of the features conformance class, and the Collections spec duplicates a subset of the Features spec. These two conformance classes can be defined in the same spec, and be clearly defined as to what they mean.

PR #331, opened 2022-09-13
incorporate collections spec into features spec

philvarner:
**Related Issue(s):** 

- https://github.com/radiantearth/stac-api-spec/issues/330

**Proposed Changes:**

1. Incorporate the Collections spec into the Features spec. This is just reorganization, and contains no semantic changes.
2. Remove Transaction Extension files, which had been inadvertently left when it was previously migrated to a separate repo

**PR Checklist:**


ogcapi-features/README.md
```diff
@@ -1,42 +1,93 @@
-# STAC API - Features
-
-- [STAC API - Features](#stac-api---features)
+# STAC API - Features and Collections Specification
+
+- [STAC API - Features and Collections Specification](#stac-api---features-and-collections-specification)
+  - [Conformance Classes](#conformance-classes)
+    - [STAC API - Features](#stac-api---features)
```
m-mohr, line 5: Doesn't this actually require STAC API - Core and thus implicitly require all the OGC API classes? Wondering whether we should just list STAC API - Core here and not require the OGC classes explicitly, otherwise we'd need to do that basically everywhere?!

ogcapi-features/README.md
```diff
@@ -1,42 +1,93 @@
...
+of OAFeat, while the STAC API - Collections (<https://api.stacspec.org/v1.0.0-rc.1/collections>) conformance
+class is the subset of Features that pertains only to Collections.
+
+### STAC API - Features
+
+The *STAC API - Features* (<https://api.stacspec.org/v1.0.0-rc.1/ogcapi-features>) conformance class
+encompasses all of the behavior described in this specification, as
+derived from OAFeat.
+
+### STAC API - Collections
```
m-mohr, line 58: Throughout the document Features is always first and then follow Collections. In my head I'd logically expect Collections first and then Features, because Collections is basically the first step you'd implement and then the additions from Features. Thoughts?

ogcapi-features/README.md
philvarner: I think we need to decide if this section means all of the dependent conformance classes also, or only the ones defined in this spec. I think it should  be just what's defined in this spec, since core is the only one that's required by any api.

Core doesn't implicitly **require** the OGC API classes -- those are optional if you want to advertise that this is also an OGC API, but you don't have to.

m-mohr: Yes, I agree that should be just what's defined in this spec. But then this doesn't need to list GeoJSON or OpenAPI, right?

philvarner: Agree, I'll remove those here and clarify the text

philvarner: clarified.

ogcapi-features/README.md
philvarner: Collections is a subset of the behavior of Features. You're not required to list collections, but you might as well do it. 

For a client that's looking to use the /collections endpoint, they should look for either one one of them.

philvarner: updated the order of Collections and Features.

philvarner:
@m-mohr ready for re-review

ogcapi-features/README.md
```diff
@@ -1,42 +1,90 @@
...
+
+The *STAC API - Collections* (<https://api.stacspec.org/v1.0.0-rc.1/collections>) conformance class
+requires only the subset of the behavior of Features that relates to Collections.
+
+This subset is:
+- the `data` link relation on the landing page
+- the `/collections` and `/collections/{collection_id}` endpoints
+
+This subset of behavior is useful in conjunction
+with the *STAC API - Item Search` conformance class to allow the description of STAC Collections
```
m-mohr, line 65: I appreciate the guidance, but I don't really understand how it is useful in conjunction
with the *STAC API - Item Search`? Why is it useful in this case?

m-mohr:
```suggestion
with the *STAC API - Item Search* conformance class to allow the description of STAC Collections
```

ogcapi-features/README.md
```diff
@@ -1,42 +1,90 @@
...
+### OGC API - Features - Part 1: GeoJSON
+
+A STAC API implementation that conforms to *STAC API - Features*
+also conforms to [OGC API - Features - Part 1 Requirements Class GeoJSON](http://docs.opengeospatial.org/is/17-069r3/17-069r3.html#_requirements_class_geojson) (<http://www.opengis.net/spec/ogcapi-features-1/1.0/conf/geojson>)
+This conformance class should be advertised to OAFeat clients so they will know of the support for GeoJSON.
+
+### OGC API - Features - Part 1: OpenAPI 3.0
+
+If OpenAPI 3.0 is used for the endpoint referenced by the `service-desc` link relation in the landing page defined
+by *STAC API - Core`, the
```
m-mohr, line 85:
```suggestion
by *STAC API - Core*, the
```

ogcapi-features/README.md
philvarner: Without this, there's no place that describes what the collections actually are. This text was already in here, so I'd prefer not to change it with this PR -- if you want to file an issue, we can review later.

ogcapi-features/README.md
```diff
@@ -49,16 +97,39 @@ The following Link relations must exist in the Landing Page (root).
 | `conformance` | `/conformance` | OAFeat   | Conformance URI     |
 | `data`        | `/collections` | OAFeat   | List of Collections |
 
+The following Link relations must exist in the `/collections` endpoint response.
```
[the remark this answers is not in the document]
philvarner: `self` is Requirement 13 https://docs.opengeospatial.org/is/17-069r4/17-069r4.html#_response_4
`root` is to be consistent with all of our other endpoints. 

If you think this should be relaxed, please file an issue.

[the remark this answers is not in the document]
m-mohr: For further discussion: #340

m-mohr (approved):

Issue #334, opened 2022-09-30
clarify meaning of limit


philvarner:
The correct behavior is what's defined in OGC API Part 1 - <http://docs.opengeospatial.org/is/17-069r3/17-069r3.html>

```
7.15.8. Error situations
...
A 400 will be returned in the following situations:
...
    If query parameter limit is not an integer or not between minimum and maximum;
```

PR #335, opened 2022-10-06
revise limit semantics

philvarner:
**Related Issue(s):** 

- #334

**Proposed Changes:**

1. limit semantics now match OAFeat. The limit parameter must be less than the maximum value advertised in the service description and the server must not return more items than the limit parameter. 

**PR Checklist:**

### DIFF
--- a/item-search/README.md
+++ b/item-search/README.md
@@ -105,8 +105,8 @@ See [examples](examples.md) for some example requests.
 
 **limit** The limit parameter follows the same semantics of the OAFeat Item resource limit parameter. The value is
 a suggestion to the server as to the maximum number of Item objects the
-client would prefer in the response. The server must return fewer Item objects, but
-may not return more. The OpenAPI specification defines the default and maximum values
+client would prefer in the response. The server may return fewer Item objects, but
+must not return more. The OpenAPI specification defines the default and maximum values
 for this parameter. The base specifications define these with a default of 10 and a maximum of 10000, but implementers
 may choose other values to advertise through their `service-desc` endpoint.  If the limit parameter value is greater
 than the advertised maximum limit, the server must return a 400 Bad Request status code.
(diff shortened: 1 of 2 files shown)


item-search/README.md
```diff
@@ -105,13 +105,13 @@ See [examples](examples.md) for some example requests.
 
 **limit** The limit parameter follows the same semantics of the OAFeat Item resource limit parameter. The value is
 a suggestion to the server as to the maximum number of Item objects the
-client would prefer in the response. The OpenAPI specification defines the default and maximum values
+client would prefer in the response. The server must return fewer Item objects, but
```
m-mohr, line 108: It is `<=` and not `<`, right?

m-mohr (changes requested):
A slight mistake in wording, otherwise LGTM.

item-search/README.md
philvarner, line 108:
```suggestion
client would prefer in the response. The server may return fewer Item objects, but
```

item-search/README.md
```diff
@@ -105,13 +105,13 @@ See [examples](examples.md) for some example requests.
...
+may not return more. The OpenAPI specification defines the default and maximum values
```
philvarner, line 109:
```suggestion
must not return more. The OpenAPI specification defines the default and maximum values
```

item-search/README.md
philvarner: fixed the wording.  it **may** return fewer, but **must not** return more.

m-mohr (approved):
LGTM. OAFeat also says:

> 

> A 400 will be returned in the following situations: If query parameter limit is **not an integer** or **not between minimum** and maximum;

Not sure whether we should cover this here though. (Also not a big fan of returning a 400 in case you are outside of the min/max bounds.)

item-search/README.md
m-mohr: Thanks.

PR #341, opened 2022-10-13
fix code samples using the http type to use none

philvarner:
**Related Issue(s):** 

- n/a

**Proposed Changes:**

1. fix code samples using the http type to use none, since none of these were http format, they showed red highlights in GH

**PR Checklist:**

### DIFF
--- a/implementation.md
+++ b/implementation.md
@@ -16,7 +16,7 @@ configuration so browser-based UIs running on a different domain may more easily
 
 APIs should acknowledge pre-flight request headers. In general, these header values should be set on responses:
 
-```http
+```
 access-control-allow-origin: *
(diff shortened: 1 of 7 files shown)


m-mohr:
What would be HTTP format?

m-mohr (approved):

philvarner:
> What would be HTTP format?

I don't actually know -- it's supposed to be for the mime type `message/http`, but I can't find an example of what that actually looks like

m-mohr:
I guess they want the full header with status codes etc...

```http
200 OK
Content-Type: text/html
```
or 
```
GET /
Host: example.com
Accept: text/html
```

PR #343, opened 2022-10-13
fix publishing schemas by renaming default from dev to main

philvarner:
**Related Issue(s):** 

- n/a

**Proposed Changes:**

1. When renaming the default branch from dev to main, I missed these two uses of dev in the schema publishing scripts. Fixing that.

**PR Checklist:**

### DIFF
--- a/.circleci/publish.js
+++ b/.circleci/publish.js
@@ -1,7 +1,7 @@
 const ghpages = require('gh-pages');
 
 let args = process.argv.slice(2);
-let tag = 'dev';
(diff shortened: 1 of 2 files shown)


matthewhanson (approved):

m-mohr:
We still have dev. Is that expected to be removed soon?

philvarner:
> We still have dev. Is that expected to be removed soon?

We haven't done a release since we switched, so once we've done that and verified everything is still working, I'll removed it.

Issue #356, opened 2022-11-04
Broken link in features spec


gadomski:
At minimum, https://github.com/radiantearth/stac-api-spec/blob/d4f8b4ab45e17e9923bd75cc58cace4819e645ff/ogcapi-features/README.md?plain=1#L22 is a broken URL. I imagine there's others, since I see similar link structures in the document.

philvarner:
Looks like the script that publishes the openapi fails with Features and Collections. Core and Item Search work and are available.

emmanuelmathot:
Same problem with many extensions referencing the https://api.stacspec.org/v1.0.0-rc.2/ogcapi-feature schema in release v1.0.0-rc.2

emmanuelmathot:
@philvarner Could you add the missing schema manually in  https://api.stacspec.org ? That would solve this issue. Thx

philvarner:
I'll see what I can do.

PR #370, opened 2023-01-13
fix conformance class publishing

philvarner:
**Related Issue(s):** 

- https://github.com/radiantearth/stac-api-spec/issues/356

**Proposed Changes:**

1. Fix conformance class schema publishing

**PR Checklist:**


m-mohr (approved):
I'm not good at bash scripting, so I just assume that you know what you are doing ;-)

Issue #362, opened 2022-11-29
Clarify parent link in Features context


m-mohr:
As far as I know the parent links in the context of STAC API - Features must always be pointing to valid Catalogs or Collections.
So the parent links must be as follows:

- parent of /collections/:collectionid/items/:itemid must be /collections/:collectionid
- parent of /collections/:collectionid/items must be /collections/:collectionid
- parent of /collections/:collectionid must be /
- parent of /collections must be /

I often see items pointing to .../items and collections to /collections. I think we should clarify this.

philvarner:
I've added documentation to clarify these two:
- parent of /collections/:collectionid must be /
- parent of /collections/:collectionid/items/:itemid must be /collections/:collectionid

I made it clear in the table that these two do not have parent links:
- parent of /collections/:collectionid/items must be /collections/:collectionid
- parent of /collections must be /

in OAFeat, /collections/:collectionid/items has a link rel `collection` so that it's not confused the the parent-child relation. We could add something similar to /collections (catalog?), but there's already root which points to the same link.

tschaub:
The item spec includes [this requirement](https://github.com/radiantearth/stac-spec/blob/v1.0.0/item-spec/item-spec.md?plain=1#L223):
> An Item must also provide a link to the STAC Collection using the `collection` relation type

Then in STAC API, the link from an item to its collection [has the `parent` relation type](https://github.com/radiantearth/stac-api-spec/blob/f7c2fb15664bd7934bd93a69ba4872a356c81d40/ogcapi-features/README.md#items-collectionscollectioniditemsitemid).

I *think* STAC Browser assumes the link from item to collection will have the `collection` relation.  Not sure what else makes this assumption.

Should this be `parent` or `collection`?

philvarner:
`collection` is the right relation if you want to definitely find the Collection. The `parent` link may link to a parent Catalog the Item is in (though in practice, none of the STAC API impls have sub-Catalog support yet, so it's just root Catalog -> Collections -> Items).

tschaub:
Thanks for the clarification, @philvarner.  I've opened #404 proposing that `collection` be used instead of `parent` for links from an item to its collection.

philvarner:
It's not instead of though -- they're frequently the same (and always the same in all the STAC API impls I know of), but they don't have to be the same -- collection will always point to the Collection resource, but parent may *not* point to the collection, but might point to a parent Catalog.

[...]

Issue #350, opened 2022-10-26
Features endpoint /collection/{c_id}/items does not state it must have a self, root, parent, and collection link


philvarner:
Features endpoint /collection/{c_id}/items does not state it must have a self, root, or parent link.

self and collection is required by OAFeat, so we just need to explicitly state it is required.

from <https://docs.opengeospatial.org/is/17-069r4/17-069r4.html#_response_6>:

A | A 200-response SHALL include the following links in the response:
- a link to the response document (relation: self),
- a link to the response document in every other media type supported by the service (relation: alternate), and 
- a link to the feature collection that contains this feature (relation: collection).

We generally recommend root is a link rel.

Parent should also be a link rel, I think?

PR #371, opened 2023-01-16
claify features links wrt to parent link

philvarner:
**Related Issue(s):** 

- https://github.com/radiantearth/stac-api-spec/issues/362
- https://github.com/radiantearth/stac-api-spec/issues/350

**Proposed Changes:**

1. Clarify what the parent link should point to
2. Add descriptions of Items endpoints

**PR Checklist:**


ogcapi-features/README.md
```diff
@@ -122,6 +133,33 @@ Usually, the `self` link in a Collection must link to the same URL that was used
...
+
+### Collection Items (/collections/{collectionId}/items)
+
+The following Link relations must exist in the ItemCollection object returned from the `/collections/{collectionId}/items` endpoint.
+
+| **rel**      | **href**                            | **Media Type**       | **From**  | **Description**      |
+| ------------ | ----------------------------------- | -------------------- | --------- | -------------------- |
+| `root`       | `/`                                 | application/json     | STAC Core | The root URI         |
+| `self`       | `/collections/{collectionId}/items` | application/geo+json | OAFeat    | Self reference       |
+| `collection` | `/collections/{collectionId}`       | application/json     | OAFeat    | Collection reference |
```
m-mohr, line 147: This one keeps me a bit puzzled. Everywhere in STAC we use parent, but here it is collection. That's not very consistent. I'm wondering whether we should require parent and collection for consistency?

ogcapi-features/README.md
```diff
@@ -97,13 +104,17 @@ The following Link relations must exist in the Landing Page (root).
 | `conformance` | `/conformance` | OAFeat   | Conformance URI     |
 | `data`        | `/collections` | OAFeat   | List of Collections |
 
+### Collections (/collections)
+
 The following Link relations must exist in the `/collections` endpoint response.
 
 | **rel** | **href**       | **From**  | **Description** |
 | ------- | -------------- | --------- | --------------- |
 | `root`  | `/`            | STAC Core | The root URI    |
```
m-mohr, line 113: This requirement can't really come from STAC Core as a Collections list does not exist in STAC Core.

ogcapi-features/README.md
philvarner: The difference is that this is a link rel from the ItemCollection rather than an Item. The use of `collection` as the link rel is actually from OAFeat, since the Collection is the parent of all the Items, but *not* the parent of the ItemCollection (which has no parent).

ogcapi-features/README.md
philvarner: Agree about From not being Core -- I'll fix that.

Same for /items -- the root Catalog is the parent of all the Collections, but not the "CollectionCollection" returned by /collections, so we don't use parent here. Presumably we could add the equivalent of the `collections` link rel in /items, but it's always going to be the same as root (I think?) and I don't know what link rel name would be -- catalog?

cholmes (approved):

m-mohr (approved):

PR #374, opened 2023-01-17
make use of status codes consistent and recommended status codes for authentication

philvarner:
**Related Issue(s):** 

- https://github.com/radiantearth/stac-api-spec/issues/354

**Proposed Changes:**

1. make definitions of status codes more consistent
2. add recommneded status code when authentication must be applied

**PR Checklist:**


core/openapi.yaml
```diff
@@ -37,7 +37,7 @@ paths:
       responses:
         '200':
           $ref: '#/components/responses/LandingPage'
-        '500':
+        default:
```
m-mohr, line 40: I'm not sure default is ideal here as it would also apply to 2XX and 3XX.

It seems you can use 4XX and 5XX instead of default:
https://swagger.io/docs/specification/describing-responses/

philvarner: Updated to use 4XX and 5XX. Funny thing -- the docs actually give an example of using `default` for the error response, which I think I copied without realizing it wasn't accurate.

m-mohr (approved):

Issue #250, opened 2022-01-10
STAC API - Browseable Specification: Differentiate between API and static


m-mohr:
As an implementor, it is confusing how to handle the STAC API - Browseable Specification. There's no hint at how a missing conformance class for the "STAC API - Browseable Specification" should be handled and how to distinguish from static STACs which don't have this conformance class.

Some wild assumptions:
- no conformsTo in root landing page -> static catalog -> show child and item links
- stac_version >= 1.0.0-beta.5 and conformsTo present but no browsable conformance class -> API -> don't show child and item links although some might be present?
- stac_version <= 1.0.0-beta.5 and conformsTo present -> API -> ???
- stac_version >= 1.0.0-beta.5 and conformsTo with browsable conformance class present -> API -> show child and item links

Or is it:
- always show available child and item links

Having this in mind, what's the advantage of adding this conformance class to my implementation? I don't really see what it adds to a client such as STAC Browser and how I should handle the different variations in the client.

The only advantage I see right now is that if it is present, in the STAC Index Crawler I can skip the OGC API - Features completely when the conformance class is given.

philvarner:
I think this is more clear than when you filed this, but I think this speaks to the value of Browseable:

```
A STAC API conforming to the *STAC API - Browseable* conformance class
(<https://api.stacspec.org/v1.0.0-rc.1/browseable>) must be structured such that
all Items in the catalog can be accessed by following `child` and `item` link relations. This is a more significant
constraint than a STAC API without this conformance class or a STAC Catalog that is available over HTTP but does not
implement STAC API, neither of which have any guarantee regarding the reachability of Items through these link relations. This conformance 
class is used to signal to users that they can fully navigate to all available Items using a UI (like [STAC Browser](https://github.com/radiantearth/stac-browser),
and also makes it clear to crawlers that they can reach everything by following catalog links.
```
In terms of what to show for a Browse interface, I think this is the decision tree:

- no conformsTo - not a STAC API
    -  child + item traversal, if every item isn't reachable then it's not very useful
      - Action: show child and item links 
- conformsTo - is a STAC API
    - Children - pageable way to traverse
      - Action: show pages of children results (Catalogs and Collections) 
    - Browseable - child + item traversal, will be able to reach every item
      - Action: show child and item links
    - Features
      - Action: heuristic on child + item traversal and /collections/{c_id}/items pagination -- this is hard to figure out, because the catalog could structure these in any way they want
    - Collections
[...]

Issue #246, opened 2022-01-05
Add diagram for browseable relations


philvarner:
There is clear text description of the various graph catalog structures that can be constructed with child and item link relations, but some diagrams of this would help clarify these.

Issue #243, opened 2021-12-16
Improve documentation around browse use cases


philvarner:
Currently, the use cases for browse are not well-understood or documented. These should be articulated, and from them, recommendations should be derived on how to structure catalog hierarchies to support these use cases.

PR #377, opened 2023-01-29
updates to browseable

philvarner:
**Related Issue(s):** 

- https://github.com/radiantearth/stac-api-spec/issues/243
- https://github.com/radiantearth/stac-api-spec/issues/250
- https://github.com/radiantearth/stac-api-spec/issues/246

**Proposed Changes:**

1. Update description as to what Browseable actually defines on an API
2. Add diagram of child-item relations
3. Clarify use cases a bit

**PR Checklist:**


core/README.md
```diff
@@ -81,16 +66,31 @@ that declare different conformance classes. This is useful when an entire catalo
...
+to advertise that all Items in the Catalog presented by a STAC API may be reached by following `child` and
+`item` link relations. In a non-API STAC Catalog, all items must be reachable via these relations to be
+considered in the catalog. With a STAC API Catalog, items are considered to be in the catalog if they are
+accessible via search operations. However, a STAC API Catalog may also have child and item link relations.
+This presents an ambiguity between whether the same set of items accessible via search operations
+is also accessible via link relations. Browseable clarifies this by unambiguously advertising that
+items available via search can also be accessed by following link relations.
+
+Support for this "browse" mode of interaction via link relations is complementary to the dynamic search
+capabilities defined in the *STAC API - Features* and *STAC API - Item Search* conformance classes.
```
cholmes, line 81: Maybe put links to the features and item searches? Just so people can easily get to them to comapare.

cholmes (approved):
Looks great, nice updates.

core/README.md
```diff
@@ -81,16 +66,31 @@ that declare different conformance classes. This is useful when an entire catalo
...
-implement STAC API, neither of which have any guarantee regarding the reachability of Items. This conformance 
-class is used to signal to users that they can fully navigate to all available Items using a UI (like [STAC Browser](https://github.com/radiantearth/stac-browser), 
-and also makes it clear to crawlers that they can reach everything by following catalog links. 
+## Browseable
+
+The Browseable conformance class (<https://api.stacspec.org/v1.0.0-rc.2/browseable>) provides a formal way
+to advertise that all Items in the Catalog presented by a STAC API may be reached by following `child` and
+`item` link relations. In a non-API STAC Catalog, all items must be reachable via these relations to be
+considered in the catalog. With a STAC API Catalog, items are considered to be in the catalog if they are
+accessible via search operations. However, a STAC API Catalog may also have child and item link relations.
```
m-mohr, line 75: I have a couple of questions:
- While I like this way of thinking, but what about APIs that don't support item search?
- Would it in this case mean that all items can be reached via item/child links that are part of STAC API - Features?
- Should Browsable be used if an API only implements STAC API - Collections and Collection Search and exposes all Collections via child links (but has no items at all)?

core/README.md
```diff
@@ -278,7 +306,7 @@ present each of these as a catalog, and create parent catalogs for them that all
       - child -> /catalogs/sentinel_5p_cloud
   - child -> /catalogs/modis
     - child -> /catalogs/modis_mcd43a4
-    - child -> /catalogs/modis_mod11a1
```
m-mohr, line 281: The mermaid stuff above is fancy. Maybe we should also use it for these examples?

core/README.md
philvarner: done

core/README.md
philvarner: Filed https://github.com/radiantearth/stac-api-spec/issues/386 for that

core/README.md
philvarner: >     * While I like this way of thinking, but what about APIs that don't support item search?

Good point. I'll clarify this language, since it should be considered in the catalog of it's accessible via link relations **or** search via /search or /items
 
>     * Would it in this case mean that all items can be reached via item/child links that are part of STAC API - Features?

Yes -- I clarified that search includes both /search and /items

>     * Should Browsable be used if an API only implements STAC API - Collections and Collection Search and exposes all Collections via child links (but has no items at all)?

Yes, I think so. I'm reluctant to add this as an example because it seems like a very uncommon case.

1
```diff
@@ -0,0 +1,20 @@
+Merge branch 'main' into pv/browseable
```
m-mohr, line 1: What's this?

philvarner: seems like a stray file that was accidentally committed -- deleted.

philvarner:
@m-mohr did you have any other issues you wanted addressed, or is this ok to merge?

m-mohr:
The PR itself is good. I have general concerns about this conformance class as part of Core as it's not implemented anywhere yet (afaik). Even the newly mentioned clients like STAC Browser, pystac-client and stac-nb don't implement it, right? I also don't really know how I could implement it in STAC Browser so that it's beneficial for users. Thoughts? What were your ideas when you mentioned it? @philvarner

philvarner:
I think that's a reasonable concern. I think originally it was part of Core instead of an Extension when all of the extensions were still in this repo, but now that they're separate, I would be in favor of separating it out into an extension that can be versioned and matured independently. This would also align with some of the concerns I have about how much "best practices" text we have in the spec, and this conformance class has a lot of explanatory text for being effectively one sentence of semantics.

Maybe we merge this PR, and then I'll move it to an extension before rc.3?

m-mohr:
@philvarner Makes sense to me. Go ahead.

Issue #239, opened 2021-12-10
Create a new link relation like "parent-catalog" or "api-parent"


philvarner:
Currently, we have two link relations that go "up" the tree -- root and parent. However, in deeply nested trees of catalogs that have varying support for conformance classes, these are not sufficient to express the "root" of a sub-Catalog. 

Example:
CMR has a root that doesn't support search (https://cmr.earthdata.nasa.gov/stac), but links via child relations to catalogs that do (https://cmr.earthdata.nasa.gov/stac/LARC_ASDC) . An item from one of these catalogs should have "root" as the top level-root, an parent may only go up one level in a tree of catalogs, so there's a need to express the "search root".

chiarch84:
Dear @philvarner I think that this [issue](https://github.com/radiantearth/stac-api-spec/issues/329#issuecomment-1260668915) is related to what you are saying here. 
I guess that, if at each level of "deepness" of the catalog tree structure, you implement the /search method (and of course all the other methods) it will be then just an implementation problem (so how to create the query) to correctly search only on that specific tree branch.

Let's create a simple example made of catalogs, subcatalogs, subsubcatalogs and collections etc.:

**Root**
--- **sub-catalog A**
------- Collection A1
-------------- Item A1.1
-------------- Item A1.2
-------------- Item ...
------- Collection A2
------- **subsub-catalog A1**
-------------- Collection A1.1
------------------ Item A1.1.1
-------------- Collection A1.2
--- **sub-catalog B**
------- Collection B1
------- Collection B2

If the APIs implemented are:
/catalog/{catalogid}
/catalog/{catalogid}/collections
/catalog/{catalogid}/collections/{collectionid}
/catalog/{catalogid}/collections/{collectionid}/items
/catalog/{catalogid}/collections/{collectionid}/items/{itemid}
/catalog/{catalogid}/search
/catalog/{catalogid}/children

**It would be like if I there were 4 different STAC API endpoints**:
{baseurl}/catalogs/root.catalog
{baseurl}/catalogs/A.catalog
{baseurl}/catalogs/A1.catalog
{baseurl}/catalogs/B.catalog

**Each of these ones should permit the /search** (and all other methods) function. 
It will be then the technical implementation that, in the case of {baseurl}/catalogs/A.catalog/search will have to search, not only in items A1.1 and A1.2, but also in item A.1.1.1. 

I hope my comment makes sense in this issue, if it is not in the context please let me know.

[...]

PR #378, opened 2023-01-29
Proposal: add concestor link relation

philvarner:
**Related Issue(s):** 

- #239 

**Proposed Changes:**

1. Add `concestor` link relation to furthest back parent

**PR Checklist:**

### DIFF
--- a/core/README.md
+++ b/core/README.md
@@ -7,14 +7,6 @@
   - [Extensions](#extensions)
   - [Structuring Catalog Hierarchies](#structuring-catalog-hierarchies)
 
-- **OpenAPI specification:** [openapi.yaml](openapi.yaml) ([rendered version](https://api.stacspec.org/v1.0.0-rc.2/core)),
-- **Conformance URIs:**
-  - <https://api.stacspec.org/v1.0.0-rc.2/core>
-  - <https://api.stacspec.org/v1.0.0-rc.2/browseable>
-- **[Maturity Classification](../README.md#maturity-classification):** Candidate
-- **Dependencies**: None
-  and [commons.yaml](commons.yaml) is the OpenAPI version of the core [STAC spec](../stac-spec) JSON Schemas.
-
 All STAC API implementations must implement the *STAC API - Core* specification. The conformance class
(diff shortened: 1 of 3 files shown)


core/README.md
```diff
@@ -101,26 +93,31 @@ OpenAPI 3.0 or 3.1 with media types `application/vnd.oai.openapi` (YAML),
...
+
+The `service-doc` link relation and endpoint is recommended, but not required. This commonly returns an HTML
 page, for example, in the form of [Redoc](https://github.com/Redocly/redoc) interactive API
 , but any format is allowed. The Link `type` field should correspond to whatever format or formats are
 supported by this endpoint, e.g., `text/html`.
 
-| **rel**       | **href**    | **From** | **Description**                                                                                                                    |
-| ------------- | ----------- | -------- | ---------------------------------------------------------------------------------------------------------------------------------- |
-| `service-doc` | `/api.html` | OAFeat   | A human-consumable service description. The path for this endpoint is only recommended to be `/api.html`, but may be another path. |
+The `concestor` link is to the most distant ancestor catalog if this catalog is in a tree of sub-catalogs.
```
m-mohr, line 110: I'm not sure I understand. First of all, `concestor` I've never heard before but I'm also a non-native spreaker. Assuming I'm somewhat "average", this might not be as ituitive as the other relation types. 

Also, from the description below I'm not sure what the difference between root and concestor is. Initially, I'd think that if I follow all parent links I'd get to the root. So this likely needs more explanation.

Would this be a good extension as it seems relatively niche?

m-mohr:
After the discussion yesterday, I think we were coming from different directions. Assuming we are looking at CMR which has a catalog structure as "real" root and a couple of APIs underneath it.
1. Point the root to the API landing pages and point concestor to the "real" root
2. Point the root to the "real" root and something like "parent-api" to the API landing pages

I'm honestly not sure what the initial desire in #239 was, maybe the involved users were actually looking at different use cases, too.
Point 2 feels more logical to me though as calling something root which is not the actual root is confusing. Also I feel it is more explicit. I'd struggle to implement 1 in STAC Browser, I think.

Issue #361, opened 2022-11-29
Conformance classes in OpenAPI schema


emmanuelmathot:
It would be useful to have the conformance classes directly specified in the schemas. We could use it to automatically generate the landing page in the case of a STAC API framework leveraging code generation.
It would be possible to specify them at schema or operation level for a fine granularity. For instance with `item-search`:

```yaml
openapi: 3.0.3
info:
  title: STAC API - Item Search
  version: 1.0.0-rc.2
  description: >-
    This is an OpenAPI definition of the SpatioTemporal Asset Catalog API - Item Search
    specification.
  contact:
    name: STAC Specification
    url: 'http://stacspec.org'
  license:
    name: Apache License 2.0
    url: 'http://www.apache.org/licenses/LICENSE-2.0'
  conformance: [ "https://api.stacspec.org/v1.0.0-rc.2/item-search" ]
tags:
  - name: Item Search
    description: essential characteristics of a STAC API
paths:
[...]
```

philvarner:
Additional fields have to start with `x-`. I'm adding a field with name `x-conformance-classes`.

philvarner:
These were added to the Foundation specs. I've made a mental note to add it to the extensions (but not filed the ~8 issues against every extension that could use it)

Issue #31, opened 2020-06-24
OpenAPI 3.1


m-mohr:
OpenAPI 3.1 is in release candidate phase: https://github.com/OAI/OpenAPI-Specification/releases/tag/3.1.0-rc0

There most important change for us is:

> The Schema Object is now fully compliant with JSON Schema draft 2019-09
  
It allows us to re-use the JSON Schemas of the core STAC repo and get rid of many lines of OpenAPI schemas, it also allows to link to GeoJSON JSON Schema, for example. Related issue it would help with are #29 and #8.

Drawback will be that tooling will take some time to adopt, especially Spectral (for validation) and ReDoc (for documentation) are important.

Related: https://github.com/opengeospatial/ogcapi-features/issues/404

cholmes:
Putting this on beta.2, since it seems like we should wait till it 3.1.0 final, and hopefully tooling can catch up a bit.

m-mohr:
Likely even quite a bit later :-)

PR #382, opened 2023-01-30
Pv/openapi

philvarner:
**Related Issue(s):** 

- https://github.com/radiantearth/stac-api-spec/issues/31
- https://github.com/radiantearth/stac-api-spec/issues/361

**Proposed Changes:**

1. Update schemas to OpenAPI 3.1
2. Clarify that the openapi30 conformance class from OAFeat shouldn't be used with these and remove from example landing pages
3. add `x-conformance-classes` extension property to `info` section

**PR Checklist:**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,9 +6,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## Changed
+
+- OpenAPI schemas are now OpenAPI 3.1 instead of OpenAPI 3.0.
+
 ## Added
 
 - Added authentication status code recommendations.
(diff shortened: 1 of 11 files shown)


m-mohr:
I know I opened #31, but as OGC API seems still to use 3.0, I'm wondering whether we should stick to 3.0 for now to now mix versions and have less issues with tooling? I think 3.1 support in tooling is still evolving and we also don't really make use of any 3.1 improvements yet (i.e. JSON Schema instead of "OpenAPI schema") so there's no urgent reason to migrate, right?

philvarner:
Okay, I put this into draft and created a new PR for just the x-conformance-classes addition https://github.com/radiantearth/stac-api-spec/pull/385

PR #404, opened 2023-03-07
Use collection relation for links to collection

tschaub:
**Related Issue(s):** 

- [#362](https://github.com/radiantearth/stac-api-spec/issues/362#issuecomment-1457397588)

**Proposed Changes:**

1. Use the `collection` relation type for links from an item to a collection

The STAC Collection spec includes [this language](https://github.com/radiantearth/stac-spec/blob/v1.0.0/collection-spec/collection-spec.md#relation-types):

> it is REQUIRED that Items linked from a Collection MUST refer back to its Collection with the collection relation type

This proposed change makes the STAC API spec consistent with that requirement.

**PR Checklist:**

- [ ] This PR does not make any changes to the core spec in the `stac-spec` directory (these are included as a subtree and should be updated directly in [radiantearth/stac-spec](https://github.com/radiantearth/stac-spec))

### DIFF
--- a/ogcapi-features/README.md
+++ b/ogcapi-features/README.md
@@ -157,8 +157,8 @@ The following Link relations must exist in the Item object returned from the `/c
 | **rel**  | **href**                                     | **Media Type**       | **From**            | **Description**                                     |
 | -------- | -------------------------------------------- | -------------------- | ------------------- | --------------------------------------------------- |
 | `root`   | `/`                                          | application/json     | STAC API - Features | The root URI                                        |
-| `collection` | `/collections/{collectionId}`            | application/json     | STAC Collection     | The containing Collection |
-| `parent` | `/collections/{collectionId}`                | application/json     | OAFeat              | Parent reference, usually the containing Collection |
+| `collection` | `/collections/{collectionId}`            | application/json     | STAC Item           | The containing Collection |
+| `parent` | `/collections/{collectionId}`                | application/json     | STAC Item           | Parent reference, usually the containing Collection |
 | `self`   | `/collections/{collectionId}/items/{itemId}` | application/geo+json | OAFeat              | Self reference                                      |
 
 The `parent` link for an Item may point to a Collection or a Catalog.  The `collection` link for an Item will always point to the containing Collection.  Links to a Collection must point to the `/collections/{collectionId}` endpoint, rather than the API sub-path of `/collections/{collectionId}/items/`.
--- a/ogcapi-features/README.md
+++ b/ogcapi-features/README.md
@@ -157,12 +157,11 @@ The following Link relations must exist in the Item object returned from the `/c
 | **rel**  | **href**                                     | **Media Type**       | **From**            | **Description**                                     |
 | -------- | -------------------------------------------- | -------------------- | ------------------- | --------------------------------------------------- |
 | `root`   | `/`                                          | application/json     | STAC API - Features | The root URI                                        |
+| `collection` | `/collections/{collectionId}`            | application/json     | STAC Collection     | The containing Collection |
 | `parent` | `/collections/{collectionId}`                | application/json     | OAFeat              | Parent reference, usually the containing Collection |
 | `self`   | `/collections/{collectionId}/items/{itemId}` | application/geo+json | OAFeat              | Self reference                                      |
 
-Note that the `parent` link for an Item should point to the containing Collection
-(e.g., `/collections/{collectionId}`), rather than the API sub-path
-of `/collections/{collectionId}/items/`.
+The `parent` link for an item may point to a Collection or a Catalog.  A `collection` link for an Item will always point to the containing Collection.  Links to a Collection must point to the `/collections/{collectionId}` endpoint, rather than the API sub-path of `/collections/{collectionId}/items/`.
 
 ## Endpoints
 
--- a/ogcapi-features/README.md
+++ b/ogcapi-features/README.md
@@ -161,7 +161,7 @@ The following Link relations must exist in the Item object returned from the `/c
 | `parent` | `/collections/{collectionId}`                | application/json     | OAFeat              | Parent reference, usually the containing Collection |
 | `self`   | `/collections/{collectionId}/items/{itemId}` | application/geo+json | OAFeat              | Self reference                                      |
 
-The `parent` link for an item may point to a Collection or a Catalog.  A `collection` link for an Item will always point to the containing Collection.  Links to a Collection must point to the `/collections/{collectionId}` endpoint, rather than the API sub-path of `/collections/{collectionId}/items/`.
+The `parent` link for an Item may point to a Collection or a Catalog.  The `collection` link for an Item will always point to the containing Collection.  Links to a Collection must point to the `/collections/{collectionId}` endpoint, rather than the API sub-path of `/collections/{collectionId}/items/`.
 
 ## Endpoints
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added extension field to all OpenAPI specifications `x-conformance-classes` indicating the
   conformance classes they define.
 - STAC API - Item Search now requires a `root` link relation in the response from `/search`
+- Added a `collection` link from and Item to its Collection to conform with the STAC spec.
 
 ### Fixed
 


m-mohr:
I think both should be used for maximum interoperability in clients. I'm pretty sure this change will break STAC Browser to some extent as it expected a complete set of parent links from item to root.

ogcapi-features/README.md
```diff
@@ -157,10 +157,10 @@ The following Link relations must exist in the Item object returned from the `/c
 | **rel**  | **href**                                     | **Media Type**       | **From**            | **Description**                                     |
 | -------- | -------------------------------------------- | -------------------- | ------------------- | --------------------------------------------------- |
 | `root`   | `/`                                          | application/json     | STAC API - Features | The root URI                                        |
-| `parent` | `/collections/{collectionId}`                | application/json     | OAFeat              | Parent reference, usually the containing Collection |
+| `collection` | `/collections/{collectionId}`            | application/json     | OAFeat              | The containing Collection |
```
philvarner, line 160: It's not "instead of" -- they're frequently the same (and always the same in all the STAC API impls I know of), but they don't have to be the same -- collection will always point to the Collection resource, but parent may *not* point to the collection, but might point to a parent Catalog.

I think we need to add to the definitions that there should be a `collection` link rel (as defined by STAC) in addition to `parent` (defined by STAC API)

This should also be added to the similar table in Item Search.

philvarner (changes requested):
comments inline

philvarner:
Please add a CHANGELOG entry per the PR template checkboxes.

tschaub:
> I'm pretty sure the proposed change would break STAC Browser to some extent as it expects a complete set of parent links from item to root.

@m-mohr - there could be a breaking change I haven't noticed, but I made this change to get STAC Browser working.  When I used the `parent` relation instead of the `collection` relation, STAC Browser was trying to navigate "up" to the `/collections/{c}/items` route instead of the `/collections/{c}` route.  Using `collection` instead of `parent` made it so STAC Browser correctly identified `/collections/{c}` as the destination for "up" and "collection" from the item.

philvarner:
That sounds like a bug in the impl-- the parent of /collection/{c_id}/items/{item_id} is /collection/{c_id}, but I've seen implementions incorrectly point `parent` link rel at /collection/{c_id}/items -- which is not allowed, because that endpoint is a transient ItemCollection rather than a Catalog or Collection.  

If you're able to share either the endpoint or the Item json, it would help to validate that.

ogcapi-features/README.md
tschaub: > I think we need to add to the definitions that there should be a collection link rel (as defined by STAC) in addition to parent (defined by STAC API)

I restored the `parent` link in the latest commit, so the table now includes both `parent` and `collection` links.  Both use the same `href`.  Let me know if you think the `parent` link should use a different example `href`.

tschaub: > This should also be added to the similar table in Item Search.

I'm not finding this table.  Let me know if you've got a link.

ogcapi-features/README.md
```diff
@@ -157,10 +157,11 @@ The following Link relations must exist in the Item object returned from the `/c
 | **rel**  | **href**                                     | **Media Type**       | **From**            | **Description**                                     |
 | -------- | -------------------------------------------- | -------------------- | ------------------- | --------------------------------------------------- |
 | `root`   | `/`                                          | application/json     | STAC API - Features | The root URI                                        |
+| `collection` | `/collections/{collectionId}`            | application/json     | OAFeat              | The containing Collection |
 | `parent` | `/collections/{collectionId}`                | application/json     | OAFeat              | Parent reference, usually the containing Collection |
 | `self`   | `/collections/{collectionId}/items/{itemId}` | application/geo+json | OAFeat              | Self reference                                      |
 
-Note that the `parent` link for an Item should point to the containing Collection
+Note that the `parent` and `collection` link for an Item should point to the containing Collection
```
philvarner, line 164: The original text for this is overly-constrained, so we should fix that here. 

This statement is true for the collection link, but the parent link may either point to the collection or to a catalog, so there needs to be a qualifier like "if the parent link is to the collection, make sure it po;ints to the collection rather than the `/items`"

It would also be good to clarify here that the parent link may be to either a collection (typical) or to a sub-catalog URI (currently unspecified by STAC API).

ogcapi-features/README.md
philvarner: 👍🏻

ogcapi-features/README.md
```diff
@@ -157,10 +157,11 @@ The following Link relations must exist in the Item object returned from the `/c
 | **rel**  | **href**                                     | **Media Type**       | **From**            | **Description**                                     |
 | -------- | -------------------------------------------- | -------------------- | ------------------- | --------------------------------------------------- |
 | `root`   | `/`                                          | application/json     | STAC API - Features | The root URI                                        |
+| `collection` | `/collections/{collectionId}`            | application/json     | OAFeat              | The containing Collection |
```
philvarner, line 160: is the `collection` link rel defined and/or required by OAFeat, or only STAC Item?

philvarner:
Also, I have no idea why the CircleCI build is failing. as long as npm check passes, I'm fine merging without it.

ogcapi-features/README.md
tschaub: The latest commit updates this language.  Let me know if that looks ok.

ogcapi-features/README.md
tschaub: The latest commit updates this to be "STAC Item" instead of "OAFeat."

I don't see "parent" in either OGC API – Features or Common.  Should the "From" cell for "parent" also be updated to point to the STAC spec?

ogcapi-features/README.md
```diff
@@ -157,12 +157,11 @@ The following Link relations must exist in the Item object returned from the `/c
 | **rel**  | **href**                                     | **Media Type**       | **From**            | **Description**                                     |
 | -------- | -------------------------------------------- | -------------------- | ------------------- | --------------------------------------------------- |
 | `root`   | `/`                                          | application/json     | STAC API - Features | The root URI                                        |
+| `collection` | `/collections/{collectionId}`            | application/json     | STAC Item           | The containing Collection |
 | `parent` | `/collections/{collectionId}`                | application/json     | OAFeat              | Parent reference, usually the containing Collection |
```
tschaub, line 161: Does the `parent` relation really come from OGC API – Features?

m-mohr:
@tschaub The navigation issue in STAC Browser sounds like an issue in the underlying data. As @philvarner mentioned, usually the links in the data are note correct then. The navigation using parent and/or collection work in valid implementations afaik. Feel free to send me link or JSON document and I can verify what's going wrong.

tschaub:
Summarizing the current state of things as I understand it.

The [STAC Item spec](https://github.com/radiantearth/stac-spec/blob/master/item-spec/item-spec.md#relation-types) describes the `parent` relation type, suggesting that a link may be present from an item to a parent collection or catalog.  I don't see wording that suggests that it is a requirement that items must have a `parent` link pointing to a collection or catalog (but I could be missing it).

The [STAC Collection spec](https://github.com/radiantearth/stac-spec/blob/master/collection-spec/collection-spec.md#relation-types) has a note that requires that items in a collection link back to the collection with the `collection` relation.  The [Item spec](https://github.com/radiantearth/stac-spec/blob/master/item-spec/item-spec.md#collections) reiterates this requirement (that an item link to a collection with the `collection` relation type).

It looks to me like the STAC API spec adds the requirement that an item in a collection also has a `parent` link pointing to the collection (in addition to the `collection` link required by the collection spec).

https://github.com/radiantearth/stac-api-spec/blob/f7c2fb15664bd7934bd93a69ba4872a356c81d40/ogcapi-features/README.md#L155-L161

So a `collection` link from an item to collection is already required by the STAC Catalog spec.  Is it intentional that the STAC API add the requirement that a `parent` link from item to collection must also be present?

ogcapi-features/README.md
```diff
@@ -157,12 +157,11 @@ The following Link relations must exist in the Item object returned from the `/c
 | **rel**  | **href**                                     | **Media Type**       | **From**            | **Description**                                     |
 | -------- | -------------------------------------------- | -------------------- | ------------------- | --------------------------------------------------- |
 | `root`   | `/`                                          | application/json     | STAC API - Features | The root URI                                        |
+| `collection` | `/collections/{collectionId}`            | application/json     | STAC Collection     | The containing Collection |
 | `parent` | `/collections/{collectionId}`                | application/json     | OAFeat              | Parent reference, usually the containing Collection |
```
philvarner, line 161:
```suggestion
| `collection` | `/collections/{collectionId}`            | application/json     | STAC Item     | The containing Collection |
| `parent` | `/collections/{collectionId}`                | application/json     | STAC Item              | Parent reference, usually the containing Collection |
```

Both come from STAC Item.

ogcapi-features/README.md
```diff
@@ -157,12 +157,11 @@ The following Link relations must exist in the Item object returned from the `/c
...
 | `self`   | `/collections/{collectionId}/items/{itemId}` | application/geo+json | OAFeat              | Self reference                                      |
 
-Note that the `parent` link for an Item should point to the containing Collection
-(e.g., `/collections/{collectionId}`), rather than the API sub-path
-of `/collections/{collectionId}/items/`.
+The `parent` link for an item may point to a Collection or a Catalog.  A `collection` link for an Item will always point to the containing Collection.  Links to a Collection must point to the `/collections/{collectionId}` endpoint, rather than the API sub-path of `/collections/{collectionId}/items/`.
```
philvarner, line 164:
```suggestion
The `parent` link for an Item may point to a Collection or a Catalog.  The `collection` link for an Item will always point to the containing Collection.  Links to a Collection must point to the `/collections/{collectionId}` endpoint, rather than the API sub-path of `/collections/{collectionId}/items`.
```

philvarner:
@tschaub I think CI wasn't running because you didn't have write access to the repo -- it's either that, or you don't have the CirceCI github app configured for your account. (Ideally, we'd move to use GH instead of CircleCI for this build, but I haven't made the time to do that)

ogcapi-features/README.md
m-mohr: Sorry, I was bit unclear about this in the call. The / at the end for non-files is usually only required in self links, but there it's very important. I would say it's also a best practice to follow with all other links, but it may not be so much of an issue there as usually you always resolve against self if present.

Here's the corresponding stac-spec PR: https://github.com/radiantearth/stac-spec/pull/1214 @philvarner

m-mohr: References: 
- https://github.com/stac-utils/pystac-client/pull/373
- https://github.com/stac-utils/pystac-client/pull/405
- https://github.com/radiantearth/stac-browser/issues/241

tschaub: @m-mohr - I think this is probably better for a dedicated ticket, but it looks to me like you are talking about a problem with incorrect relative links rather than a problem with trailing slashes.

If the `self` link is `https://example.com/some/resource` and the URL for a related resource is `https://example.com/some/resource/related`, then a correct relative URL is `./resource/related`.

I feel like requiring that URLs end with a slash may be overdoing it.  And this will lead to inconsistencies with OGC API specs (OGC API – Features uses URLs like `collections/{collectionId}`, `collections/{collectionId}/items`, `collections/{collectionId}/items/{itemId}` etc.).

I know that it can be easy to get relative URLs wrong, and that resolving absolute URLs can be tricky.  But both can be done correctly without requiring that URLs end with a slash.

My personal opinion is that the OGC APIs and perhaps STAC are a bit **over**specified in requiring both specific links and specific URLs.  And I think things would be improved if OGC APIs allowed trailing slashes (it can be easier to stand up static sites that work with trailing slashes).  But I know it is too late for that.

ogcapi-features/README.md
tschaub: Updated in the latest commit.

tschaub:
I left a [similar comment above](https://github.com/radiantearth/stac-api-spec/pull/404#issuecomment-1459000488), but I'll ask again here in case it wasn't clear:

Is it intentional that the STAC API spec adds a REQUIREMENT that and Item have a `parent` link?

The STAC Item spec does not require that an Item must have a `parent` link.

Currently, the STAC API spec uses this language: https://github.com/radiantearth/stac-api-spec/blob/f7c2fb15664bd7934bd93a69ba4872a356c81d40/ogcapi-features/README.md#L155

And the following table includes a `parent` link.  The "must exist" language makes it sound like this is a requirement (not in STAC Item, only in STAC API) that an Item must have a `parent` link.

ogcapi-features/README.md
m-mohr: @tschaub Indeed, I just posted it here because we stumbled across it in this PR during the STAC community call.
Whether it's an invalid self link or relative link is not so important I think, the main issue is that it resolves incorrectly. But indeed we should maybe also update the stac-spec to mention that you can also solve this by changing the relative links. It's just simpler to add a slash then update 20 relative links in links and assets...

> My personal opinion is that the OGC APIs and perhaps STAC are a bit overspecified in requiring both specific links and specific URLs. And I think things would be improved if OGC APIs allowed trailing slashes (it can be easier to stand up static sites that work with trailing slashes). But I know it is too late for that.

I'm not sure... is it really specified in OGC APIs that trailing slashes are not allowed?

tschaub: > But indeed we should maybe also update the stac-spec to mention that you can also solve this by changing the relative links. It's just simpler to add a slash then update 20 relative links in links and assets

I think I'm missing context.  It sounds like you are starting with a problem that I don't know about.  Does the spec have language or examples in it that create this problem?

> is it really specified in OGC APIs that trailing slashes are not allowed?

No, the OGC specs do not forbid the use of trailing slashes.  But they do require that services respond (with a 200) to requests without trailing slashes.  For example, in the OGC API – Features spec, requirements 18 and 19 essentially say "requests to `/collections/{collectionId}` must respond with a 200 that includes collection metadata."  That doesn't mean that they cannot also respond to `/collections/{collectionId}/` (with a trailing slash) – but as you know the content of the response in those two cases needs to be different if (dot) relative URLs are used in links.

Again, I might not be seeing the problem that you are starting with, but I think if the STAC specs say anything about all this, it should be advice about being careful with (dot) relative URLs.  I don't see trailing slashes as magically solving all problems resolving relative URLs.  And I also don't see this as just a problem in resolving relative URLs against a "self" link.  Some implementations may use the "self" link as the base, but it also seems reasonable that a client might use a request URL as a base and ignore the "self" link when resolving relative URLs (assuming that links in the response will be relative to the request).  So I think that any best practice advice given in the spec should say that care needs to be taken to ensure that *all* links (self and other related links) are correctly formatted.

m-mohr: => https://github.com/radiantearth/stac-spec/pull/1217

philvarner:
I filed this issue: https://github.com/radiantearth/stac-api-spec/issues/406

I added that section since 1.0.0-rc.2, following the pattern of all of the other link rels that require `parent`. However, I don't think any of them should require parent, and just align with the STAC object definitions. I'll make that change. 

I was going to say that the `collection` link rel should be required because the `collection` field is required, but I can't actually find anywhere in the STAC API spec that we say it does, and I was under the mistaken impression that it was always required because it's always defined in practice in a STAC API. I think we should make this a constraint.

philvarner:
tl;dr @tschaub please merge this, and I'll handle the change to make `parent` link rels not required in another ticket.

philvarner (approved):

Issue #406, opened 2023-03-13
parent link is not required for any object


philvarner:
In *Features*, the Items (/collections/{collectionId}/items/{itemId}) table states that "The following Link relations must exist in the Item object..." and has a row for `parent`. However, this is not the case. This was added by @philvarner after 1.0.0-rc.2 release, and just following the pattern established for the other link relations. 

For all types (Catalog, Collection, Item), the parent link should be recommended, but not required.

PR #407, opened 2023-03-14
refactor link relation tables, make parent not a required field

philvarner:
**Related Issue(s):** 

- #406 

**Proposed Changes:**

1. `parent` link rel is not required for Collection or Item. In <= 1.0.0-rc.2, it was required for /collections/{collectionId}, and the requirement on /collections/{collectionId}/items/{itemId} was added since 1.0.0-rc.2. This aligns with it not being required in STAC Item and Collection definitions. 
2. Link relation tables are refactored to align with the style in stac-spec of using **REQUIRED** for required relations.

**PR Checklist:**


ogcapi-features/README.md
```diff
@@ -106,37 +106,32 @@ If the target of a Link's `type` is unknown, `type` SHOULD be set to `applicatio
...
-The following Link relations must exist in the `/collections` endpoint response.
+The following Link relations are defined for the `/collections` endpoint response.
 
-| **rel** | **href**       | **Media Type**   | **From**                                    | **Description** |
-| ------- | -------------- | ---------------- | ------------------------------------------- | --------------- |
-| `root`  | `/`            | application/json | STAC API - Features, STAC API - Collections | The root URI    |
-| `self`  | `/collections` | application/json | OAFeat                                      | Self reference  |
+| **rel** | **href**       | **Media Type**   | **From**                                                 | **Description**             |
+| ------- | -------------- | ---------------- | -------------------------------------------------------- | --------------------------- |
+| `root`  | `/`            | application/json | **REQUIRED** STAC API - Features, STAC API - Collections | The root URI                |
```
m-mohr, line 122:
```suggestion
| `root`  | `/`            | application/json | STAC API - Features, STAC API - Collections | **REQUIRED** The root URI                |
```

ogcapi-features/README.md
```diff
@@ -106,37 +106,32 @@ If the target of a Link's `type` is unknown, `type` SHOULD be set to `applicatio
...
+| `self`      | `/collections/{collectionId}` | application/json | OAFeat                                      | **REQUIRED** Self reference                                                                                                                                                                                                                                                                             |
+| `parent`    | `/`                           | application/json | OAFeat                                      | Parent reference, usually the root Catalog                                                                                                                                                                                                                                                              |
```
tschaub, line 133:
```suggestion
| `parent`    | `/`                           | application/json | STAC Item                                   | Parent reference, usually the root Catalog                                                                                                                                                                                                                                                              |
```

OGC API specs don't mention `parent` as far as I can tell (http://docs.opengeospatial.org/DRAFTS/17-069r5.html#_link_relations).

ogcapi-features/README.md
```diff
@@ -147,26 +142,29 @@ of that Collection, rather than the API sub-path of `/collections`.
...
 
-| **rel**      | **href**                            | **Media Type**       | **From**            | **Description**      |
-| ------------ | ----------------------------------- | -------------------- | ------------------- | -------------------- |
-| `root`       | `/`                                 | application/json     | STAC API - Features | The root URI         |
-| `self`       | `/collections/{collectionId}/items` | application/geo+json | OAFeat              | Self reference       |
-| `collection` | `/collections/{collectionId}`       | application/json     | OAFeat              | Collection reference |
+| **rel**      | **href**                            | **Media Type**       | **From**                         | **Description**                   |
+| ------------ | ----------------------------------- | -------------------- | -------------------------------- | --------------------------------- |
+| `root`       | `/`                                 | application/json     | **REQUIRED** STAC API - Features | The root URI                      |
```
tschaub, line 149:
```suggestion
| `root`       | `/`                                 | application/json     | STAC API - Features | **REQUIRED** The root URI                      |
```

tschaub:
Looks great, @philvarner.  Thanks for fixing this up.

m-mohr:
I'm actually not sure whether we should drop the requirement. I feel like requiring the parent links (and potentially even more) would help in the long run. I think adding links in API implementations is much "cheaper" than in static catalogs so I'd think we could require more in the API compared to static catalogs.

tschaub:
Maybe conditional requirements are too sloppy, but how about only requiring `parent` if an Item is not part of a Collection?  I've only worked with Items whose "parent" is a Collection, so I'm not sure how common Collection-less Items are in STAC API implementations.  It just struck me as odd to require that `/collections/{collectionId}/items/{itemId}` link to `/collections/{collectionId}` twice (once as `parent` and once as `collection`).

m-mohr:
That's weird as you'd need a condition in implementation, but I also don't think having them twice is an issue.
The same may occur in the collection where root and parent are pointing to the same thing. 
It's pretty common in static STAC (although not required) so I feel we could at least recommend it?

ogcapi-features/README.md
philvarner: correct, that's a pre-existing error -- glad we can fix it now.

matthewhanson (approved):

PR #441, opened 2023-12-08
Added the missing `assets` property to Collections

m-mohr:
**Proposed Changes:**

1. Added the missing `assets` property to Collections

**PR Checklist:**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,8 @@ None
 ### Changed
 
 - Field `type` is no longer required for all Link objects, but is instead strongly
+  recommended. This was added as a requirement in 1.0.0-rc.3 based on a mis-reading
+  of the OGC Features spec, and is now removed as a requirement.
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -5,29 +5,34 @@ All notable changes to this project will be documented in this file.
...
 
 None
 
 ## [v1.0.0-rc.4] - 2023-04-13
 
-## Changed
+### Changed
 
 - Field `type` is no longer required for all Link objects, but is instead strongly
-  recommended. This was added as a requirement in 1.0.0-rc.3 based on a mis-reading
```
philvarner, line 17: seems to have been removed inadvertently?

m-mohr:
Sure, ideally we align the descriptions between API and stac-spec so that they both express that.

The "as a while" in there is confusing to me, the sentence also makes sense without it?!

philvarner (approved):

PR #449, opened 2024-07-05
Update examples.md to include 'last page' response example

mhogeweg:
added 'last page' response example

**Related Issue(s):** 

- #

**Proposed Changes:**

1. 
2. 

**PR Checklist:**

### DIFF
--- a/item-search/examples.md
+++ b/item-search/examples.md
@@ -62,7 +62,9 @@ Response with `200 OK`:
 ```json
 {
     "type": "FeatureCollection",
-    "features": [],
+    "features": [
+       ...
+    ],
     "links": [
         {
             "rel": "first",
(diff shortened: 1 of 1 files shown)


item-search/examples.md
```diff
@@ -50,6 +50,41 @@ Response with `200 OK`:
...
+```
+HTTP GET /search?bbox=-110,39.5,-105,40.5
+```
+
+Response with `200 OK`:
+
+```json
+{
+    "type": "FeatureCollection",
+    "features": [],
```
m-mohr, line 65: If the resultset is empty, the previous page should've been the last page?!

mhogeweg: updated the example to include [ ... ] to indicate a non-empty result.

m-mohr: Thanks.

m-mohr (approved):

PR #450, opened 2024-07-05
Update README.md to include sentence of how clients shall interpret last page

mhogeweg:
added sentence about how to handle first/last pages by omitting the corresponding link relations in the result.

**Related Issue(s):** 

- #

**Proposed Changes:**

1. 
2. 

**PR Checklist:**

- [ ] I have added my changes to the [CHANGELOG](https://github.com/radiantearth/stac-api-spec/blob/main/CHANGELOG.md) **or** a CHANGELOG entry is not required.

### DIFF
--- a/item-search/README.md
+++ b/item-search/README.md
@@ -200,6 +200,8 @@ The href may contain any arbitrary URL parameter:
 Implementations may also add link relations `prev`, `first`, and `last`, though these are not required and may
 be infeasible to implement in some data stores.
 
+Clients shall interpre the absence of `prev` or `last` link relations to indicate the response represents respectively the first and last page of results.
+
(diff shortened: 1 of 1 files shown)


item-search/README.md
```diff
@@ -200,6 +200,8 @@ The href may contain any arbitrary URL parameter:
 Implementations may also add link relations `prev`, `first`, and `last`, though these are not required and may
 be infeasible to implement in some data stores.
 
+Clients shall interpre the absence of `prev` or `last` link relations to indicate the response represents respectively the first and last page of results.
```
m-mohr, line 203: As prev is not required, we can't interpret it as first page.

```suggestion
Clients shall interpret the absence of a link with the relation type `next` as an indication that the response represents the last page of results.
```

m-mohr (approved):

PR #464, opened 2025-11-26
Migrate to GH Actions, fix/update dependencies

m-mohr:
**Related Issue(s):** 

Closes #457, #456, #455, #454

**Proposed Changes:**

1. Update dependencies in package-lock.json
2. Update package-lock version (i.e. is generated with a newer version of npm)
3. Fix redoc version in the template (as already done in gh-pages), the spec documents were broken and didn't render anymore
4. Migrate from CircleCI (fails always) to GitHub Actions

Not sure how to fully disable circleci yet. Maybe this gets only picked up after merge?

**PR Checklist:**


.github/workflows/publish.yml
```diff
@@ -0,0 +1,90 @@
...
+  validate:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Setup Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: '16.14'
```
gadomski: Any reason this is a different node version than the other actions? If we wanted a single source of truth, we could use a `node-version-file`: https://github.com/actions/setup-node?tab=readme-ov-file#usage

.github/publish.js
gadomski, line 1: I usually use https://github.com/actions/deploy-pages, which would take care of a lot of this for us. It's ok to keep doing this `publish.js` workflow if we want to reduce the number of changes in this PR, too.

gadomski (approved):

jsignell (approved):

.github/workflows/publish.yml
m-mohr: Good point, updated.

.github/publish.js
m-mohr: Seems a bit more difficult than a pure replacement, so keeping this out of scope for this PR.

PR #479, opened 2026-08-06
Add new conformance class that requires collections parameter in Item Search #478

### LINKED ISSUE #478: /search: Multiple deployments require collections
We have at least two deployments in the wild, that can probably be considered being in the top deployments with regards to hosted data size. Both of them require the collection ID for global item search:

- Copernicus Data Space Ecosystem
- Microsoft Planetary Computer

Should we add a minimal conformance class for those so that clients can pick them up and require collections to be set?
Ideally, the deployments would not require collections to be set, but it seems for larger deployments that's what providers opt for.
Keeping it undefined just makes it to a problem of a user, who always runs into an error first and then has to find collection IDs, making the global search somewhat limited and frustrating to use.

m-mohr:
**Related Issue(s):** 

Closes #478

**Proposed Changes:**

1. Add new conformance class that requires collections parameter in Item Search

**PR Checklist:**

### DIFF
--- a/item-search/README.md
+++ b/item-search/README.md
@@ -8,6 +8,11 @@
   - [Query Parameters and Fields](#query-parameters-and-fields)
     - [Query Examples](#query-examples)
     - [Query Parameter Table](#query-parameter-table)
+      - [limit](#limit)
+      - [datetime](#datetime)
+      - [bbox](#bbox)
+      - [intersects](#intersects)
+      - [collections](#collections)
   - [Response](#response)
     - [Pagination](#pagination)
   - [HTTP Request Methods and Content Types](#http-request-methods-and-content-types)
@@ -22,6 +27,7 @@
 - **OpenAPI specification:** [openapi.yaml](openapi.yaml) ([rendered version](https://api.stacspec.org/v1.0.0/item-search))
 - **Conformance URIs:**
   - <https://api.stacspec.org/v1.0.0/item-search>
+  - <https://api.stacspec.org/v1.0.0/item-search#require-collections> (optional, see [collections](#collections))
 - **[Maturity Classification](../README.md#maturity-classification):** Candidate
 - **Dependencies**: [STAC API - Core](../core)
 - **Examples**: [examples.md](examples.md)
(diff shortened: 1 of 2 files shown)


emmanuelmathot (approved):
That's a great feature

item-search/README.md
```diff
@@ -22,6 +27,7 @@
 - **OpenAPI specification:** [openapi.yaml](openapi.yaml) ([rendered version](https://api.stacspec.org/v1.0.0/item-search))
 - **Conformance URIs:**
   - <https://api.stacspec.org/v1.0.0/item-search>
+  - <https://api.stacspec.org/v1.0.0/item-search#require-collections> (optional, see [collections](#collections))
```
gadomski, line 30: While it's a bit more restrictive, it's not a big leap from "disallow collection-less searches" to "disallow multi-collection searches". Multi-collection searches are problematic for at least two reasons:

- Items returned from multi-collection search are (usually) heterogenous, so they often require "binning" into per-collection aggregates before use by the client
- A search against every collection in the catalog is the same as a collection-less search

Would it be reasonable to add a second conformance class, `require-single-collection`?

m-mohr: Requiring a single collection doesn't really make sense to me. Then it's pretty much like using /collections/:id/items. At this point you could just disable /search completely. The proposal above just mirrors what at least two APIs do.
Additionally, do we have evidence of any implementation doing this?

gadomski: >Then it's pretty much like using /collections/:id/items

Except you can't `POST`?

>Also, do we have evidence of any implementation doing this?

No, but as we're exploring federated STAC APIs that forward search requests to upstream APIs, we'd like the ability to disallow cross-catalog searches that would require complicated item interleaving at the server level.

m-mohr: Yeah, although I have on my to do list to propose QUERY for all search related endpoints anyway soon. That would remove the POST conflicts and then you'd have that for all endpoints anyway. 

Thus, I'd propose to just mirror what is currently implemented and we can still add an additional conformance class for other usecases later.

gadomski (approved):

ghidalgo3:
@m-mohr  I would like to start including this conformance class in Planetary Computer.

m-mohr:
@ghidalgo3 Go ahead. I have to fix the CI before I can merge and I didn't get to it yet, but the conformance class is fixed so doesn't really depend on a merge or release.

